PR #1, opened 2018-08-27
Vagrant staging

bwhicks:
This pull request:

* Refactors variables to use a nested hierarchy
* Adds a staging Vagrantfile and a deploy for it
* Produces stable staging and QA deploys
(Note: when running staging playbook add `-e ref=develop`, because staging Winthrop is currently aimed at deploying current code base)


rlskoeser:
Finally got a chance to take this for a spin.

Documentation seems to be pretty thorough and clear so far.  Was able to provision a vagrant machine and get started. I had to change permissions on vagrant/id_rsa before ansible would run (permissions were too open).  Is there anyway that can be done automatically?

Tried running the winthrop staging deploy without any options - failed do to a python dependency that isn't available anymore (outdated python3 branch of piffle that no longer exists). I see your note now that you don't expect this to work anyway, but wanted to comment that the error message is pretty hard to read. Can we output the actual error message with line breaks rendered? (This is not urgent, obviously, but it would be nice to have.)

I tried again with `-e ref=develop` and this time I got an error on running `compress`. 
errored running the manage.py compress step - I think it's saying that node-sass is not installed? Again, error messages are difficult to read.  Happy to troubleshoot this myself, but not sure where to start.

Minor quibble on the naming: **winthrop_stag** is a little weird/distracting since stag is actually a word. I would recommend shorting to a non word (stg? not great either) or just using the full word (it's not that much longer).

I was looking at three variant playbooks for winthrop and wondering if we can put playbooks in sub .directories so they'll be more organized?  (It seems like it will quickly get unwieldy as we add other projects). Or maybe production could stay top-level but staging and qa could be in a winthrop folder?

I found this github issues discussing playbooks in subdirectories: https://github.com/ansible/ansible/issues/12862 
It looks like adding `roles_path = roles` to ansible.cfg gets a good portion of the way there, but there are other relative paths (like template dir) that would need an ansible home directory. Not sure what other implications it would have.

group_vars/all/vars.yml
```diff
@@ -1,3 +1,23 @@
 ---
 # The template directory (as a relative path from the top of repository)
 templates_dir: 'templates/'
+# standard variable for all repositories that use github
```
rlskoeser: This looks much better, glad you were able to move so many things into the common variables.

group_vars/winthrop_prod/vars.yml
```diff
@@ -0,0 +1,24 @@
+---
+
+# These variables are the stock configurations that either alias variables
+# from the vault or provide paths that the project itself uses.
+# vault variables first
+secret_key: '{{ vault_secret_key }}'
```
rlskoeser: Could some of these also go in common variables? ansible will still pick up the correct vault file, right?

templates/winthrop_staging/local_settings.py
```diff
@@ -0,0 +1,119 @@
...
+	'HOST': '{{ db_host }}',
+	'PORT': '3306',
+	'OPTIONS': {
+		'init_command': "SET sql_mode='STRICT_TRANS_TABLES'",
+	}
+    }
+}
+
+SOLR_CONNECTIONS = {
+    'default': {
```
rlskoeser: Shouldn't these be variables?

templates/winthrop_staging/local_settings.py
```diff
@@ -0,0 +1,119 @@
...
+        'ADMIN_URL': 'http://127.0.0.1:8983/solr/admin/cores'
+    },
+   'test': {
+        'COLLECTION': 'winthrop-test',
+        'URL': 'http://127.0.0.1:8983/solr/',
+        'ADMIN_URL': 'http://127.0.0.1:8983/solr/admin/cores'
+    }
+}
+
+CAS_SERVER_URL = 'https://fed.princeton.edu/cas/'
```
rlskoeser: Can we also make CAS & ldap info common variables?

templates/winthrop_staging/local_settings.py
```diff
@@ -0,0 +1,119 @@
+# This file is exec'd from settings.py, so it has access to and can
```
rlskoeser: Can templates be shared or included? i.e. would it make sense to have common localsettings snippets that are reused?

group_vars/winthrop_prod/vars.yml
bwhicks: Secret keys are actually unique to each instance between QA and PROD (or should be... in theory staging could piggy pack off QA), but deliberately we don't use the same secret keys so that if we have an error, we can limit damage.

templates/winthrop_staging/local_settings.py
bwhicks: I considered that approach. I think looking at a similar situation with Derrida, I will do just that. I wanted to feel my way around this one first.

templates/winthrop_staging/local_settings.py
bwhicks: I hadn't extracted them because there wasn't implicitly a reason they couldn't just appear in the template. LDAP info may be a bit of a trick because it's multiline. I will certainly try.

templates/winthrop_staging/local_settings.py
```diff
@@ -0,0 +1,119 @@
...
+# modify all the variables in settings.py.
+# If this file is changed in development, the development server will
+# have to be manually restarted because changes will not be noticed
+# immediately.
+
+
+# Add Google Analytics
+INCLUDE_ANALYTICS = True
```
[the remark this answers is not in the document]
bwhicks: Maybe. I don't think we have a unified system on this yet (winthrop and derrida share that, I believe.) Let me check and see.

templates/winthrop_staging/local_settings.py
bwhicks: Yes, they can be, and that might be a decent way to get around variables for multiline stanzas, now that you mention it.

bwhicks:
@rlskoeser I'll double-check the winthrop staging. It's been running cleanly for me, but it may be the case that I missed something. The key I thought I had added some help for but again, working in a different branch so may have done something with it. It's a pain because git stores files with 644 and 755 as its only two defaults and we need 600 even though it's a key we know to be completely insecure.

The naming of `winthrop_stag` I'll drop to `winthrop_stg` or similar. The actual issue is that I'm technically creating an inline group and so it overlaps with the real `winthrop_staging`. If I just make staging `localhost`, ansible gets confused and proceeds to try to use the actual host not the Vagrant instance. 

Re: subdirectories for playbooks, I like the idea but think given the discussion on the linked issue that we should leave them top dir. Its clearly an idea that Ansible needs to account for (and linking our issue to theirs is a good start to adding to people asking for it), and it would make a huge amount of sense... but for now, I'm inclined to think that forcing Ansible into a configuration it isn't quite intended for would be a bad idea.

templates/winthrop_staging/local_settings.py
bwhicks: Templated these out (vs a multiline variable). They can live in a .py extension file so that we don't have to worry about any jinja2 gotchas (the only thing I don't like is that I had to hard code `templates` as a path but that's likely to be 100% uniform)

group_vars/winthrop_prod/vars.yml
bwhicks: However, some of the others are common as I realized while working on Derrida. Will adjust accordingly.

bwhicks:
@rlskoeser I still can't reproduce the staging error so that may be something to look at in person. The only thing I can think of is that something didn't provision correctly on the VM (`vagrant destroy && vagrant up` would do a full purge).

I believe I've moved things so that the insecure `deploy` key will get its permissions set rightly before ever being invoked.

PR #6, opened 2018-09-10
QA group

bwhicks:
* Factors in a QA variable group and puts common solr configs in that group


bwhicks:
I had been reluctant in case `solr_admin_url` and `solr_url` varied, but I don't see any sense that will be the case, so they're now built via jina2 variables.

Thank you for catching the staging error, absolutely needed to be defined in staging group since the production value isn't defined.

PR #8, opened 2018-12-19
Slack notifications and GitHub deployments

bwhicks:
This pull request:

- [ ] Adds GitHub deployment roles `create_deployment` and `close_deployment`
- [ ] Inserts error handing into all tasks that enables the deployment to report an error on task failure and then bail out as normal using `block` and `rescue` and an included task from `create_deployment` role
- [ ] Configures mep and ppa qa to use the new roles as examples.
- [ ] Documents environmental variable or commandline options for GitHub token.

Intended functionality:

1. Ansible creates a deployment in the name of the person who owns the GitHub token
2. Notifies that deploy is pending by settings its status using the GitHub API
3. Closes the deployment as a success if successful, and in all other cases, sets it as error.

### DIFF
--- a/group_vars/all/vars.yml
+++ b/group_vars/all/vars.yml
@@ -1,4 +1,8 @@
 ---
+
+# GitHub token
+github_token: '{{ lookup("env", "ANSIBLE_GITHUB_TOKEN") | default("") }}'
+
 # email_host_password
 email_host_password: '{{ vault_email_host_password }}'
 
--- a/group_vars/all/vars.yml
+++ b/group_vars/all/vars.yml
@@ -1,7 +1,11 @@
 ---
 
-# GitHub token
+# GitHub settings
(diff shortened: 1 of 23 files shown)


group_vars/all/vars.yml
```diff
@@ -1,4 +1,12 @@
 ---
+
+# GitHub settings
+github_token: '{{ lookup("env", "ANSIBLE_GITHUB_TOKEN") | default("") }}'
+api_base: https://api.github.com
+deployments_endpoint: '{{api_base}}/repos/{{ repo }}/deployments'
+deploy_environment: '{{ staging|default(None) or qa|default(None) or production }}'
+deploy_description: 'Deploying {{ gitref }} to {{ deploy_environment }}'
```
rlskoeser: Would it be hard to also include hostname here?

bwhicks: As nearly as I can tell, those don't actually get used for the Slack notification. But I can include it anyway, since it would be a record for us that we could query if we needed to do so.

PR #10, opened 2019-02-04
Add playbook to revert previous deploys

bwhicks:
This adds a revert_last playbook that can be used to revert a last deploy.

Major question: should this update the GitHub deployment info with a new deploy? I am not 100% on that question.

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,14 +76,14 @@ ansible-playbook -e ref=GITREF name_of_playbook.yml
 The playbook will run, noting success and failures. The `-v` flag adjusts verbosity
 (adding more `v`s will produce more verbosity. Debug tasks are usually written at `2`)
 
-## Reverst last deploy
+## Revert last deploy
 
-To revert to previous deploy (which toggles the two), run, e.g.:
+To revert to previous deploy run, e.g.:
 
 ```{bash}
 ansible-playbook -e host_group=mep_qa revert_deployment.yml
 ```
-If you run this a subsequent time, it should revert itself seamlessly.
+This is not idempotent. Reversing it will require manual intervention.
 
 ## Overrides
 
--- a/README.md
+++ b/README.md
@@ -56,7 +56,7 @@ for secrets in a `local_settings.py`, but prevents worst case scenarios from
 emerging.
 
 
-### Running a playbook
+## Running a playbook
 
 To run a playbook, from your virtual environment, simply invoke:
 
@@ -76,6 +76,14 @@ ansible-playbook -e ref=GITREF name_of_playbook.yml
 The playbook will run, noting success and failures. The `-v` flag adjusts verbosity
 (adding more `v`s will produce more verbosity. Debug tasks are usually written at `2`)
 
+## Reverst last deploy
+
+To revert to previous deploy (which toggles the two), run, e.g.:
+
+```{bash}
+ansible-playbook -e host_group=mep_qa revert_deployment.yml
+```
+If you run this a subsequent time, it should revert itself seamlessly.
 
 ## Overrides
 
--- a/README.md
+++ b/README.md
@@ -78,12 +78,13 @@ The playbook will run, noting success and failures. The `-v` flag adjusts verbos
 
 ## Revert last deploy
 
-To revert to previous deploy run, e.g.:
+To revert to previous deploy run call the `revert_deployment`
+playbook with a `host_group` matching the deploy you want to
+revert, e.g.:
 
 ```{bash}
 ansible-playbook -e host_group=mep_qa revert_deployment.yml
 ```
-This is not idempotent. Reversing it will require manual intervention.
 
 ## Overrides
 
--- a/README.md
+++ b/README.md
@@ -78,12 +78,12 @@ The playbook will run, noting success and failures. The `-v` flag adjusts verbos
 
 ## Revert last deploy
 
-To revert to previous deploy run call the `revert_deployment`
+To revert to previous deploy run call the `revert_deploy`
 playbook with a `host_group` matching the deploy you want to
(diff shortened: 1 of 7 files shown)


bwhicks:
@rlskoeser This is ready to go and rebased on master. The it now simply makes `current` and the live link in `/var/www` into `previous`, with no toggle functionality.

revert_deploy.yml
```diff
@@ -0,0 +1,55 @@
...
+  environment:
+    PATH: '{{ path }}'
+    LD_LIBRARY_PATH: '{{ ld_library_path }}'
+    PYTHONPATH: '{{ python_path }}'
+  tasks:
+  - name: Pause and let user make sure command looks right.
+    pause:
+      prompt: >
+               You are about to revert {{ host_group | upper }}.
+               Press enter to proceed or ctrl+c then a to abort.
```
rlskoeser: Do you think we really need confirmation when it's a separate command to run revert? Seems like it would be unlikely to happen accidentally.

revert_deploy.yml
```diff
@@ -0,0 +1,55 @@
...
+  - name: Check that previous symlink exists.
+    stat:
+      path: "{{ install_root }}/previous"
+    register: previous
+  - fail:
+      msg: "Previous symlink does not exist."
```
rlskoeser: Suggested wording: no previous deploy available (or something more like that)

(User shouldn't care about implementation details like symlink)

revert_deploy.yml
```diff
@@ -0,0 +1,55 @@
...
+    when: previous.stat.islnk is not defined
+  - name: Get current symlink path
+    stat:
+      path: "{{ install_root }}/current"
+    register: current
+  - fail:
+      msg: "Current symlink does not exist."
```
rlskoeser: Does it matter if there is no `current` ? I can't imagine a scenario where we'd have a previous and no current, but it seem that from the perspective of reverting a deploy we wouldn't care whether or not there is a current deploy.

revert_deploy.yml
```diff
@@ -0,0 +1,55 @@
...
+    when: current.stat.islnk is not defined
+  - name: Set previous to current
```
rlskoeser: As discussed, `previous` symlink should be removed after revert to avoid confusion.

README.md
```diff
@@ -76,6 +76,14 @@ ansible-playbook -e ref=GITREF name_of_playbook.yml
 The playbook will run, noting success and failures. The `-v` flag adjusts verbosity
 (adding more `v`s will produce more verbosity. Debug tasks are usually written at `2`)
 
+## Reverst last deploy
+
+To revert to previous deploy (which toggles the two), run, e.g.:
```
rlskoeser: Documentation needs to be updated to reflect the change in logic we agreed on.

README.md
```diff
@@ -76,6 +76,14 @@ ansible-playbook -e ref=GITREF name_of_playbook.yml
 The playbook will run, noting success and failures. The `-v` flag adjusts verbosity
 (adding more `v`s will produce more verbosity. Debug tasks are usually written at `2`)
 
+## Reverst last deploy
```
rlskoeser: Typo or extra s here

revert_deploy.yml
bwhicks: I think my innate caution with anything that could touch prod was at play here. My concern is someone inadvertently specifying say `mep` when they meant `mep_qa` (for example). If you're comfortable removing it, I'm comfortable removing it, just knowing we need to aim well.

bwhicks:
@rlskoeser So, to grab questions:

1) If we want Github environments to show that we correctly have the old deployment reverted, we probably want to create a GitHub deployment and note the change. That way we document we have the old code. If we suspect this is more of an emergency measure that will be superseded by a later deploy, it's superfluous.

2) This should be testable on the staging VM via doing a deploy on two different references and then running the revert. Also a good prompt for me to double-check that everything is clean on the testing VM. Infrastructure code rots like anything else...

revert_deploy.yml
rlskoeser: It seems less likely to me that we're going to revert qa very often - since we're usually ok with it being broken (at least briefly), and probably going to investigate and figure out a proper fix anyway.  I don't think we need a secondary confirmation.

@thatbudakguy any opinion?

rlskoeser:
@bwhicks thanks for pointing out the GitHub environment concern; I expect reverting to be an emergency measure, but we might not always be able to fix immediately, depending on the problem. If it's not too difficult to update the GitHub deploy, I think that would be valuable (although that seems like it could make it a lot more complicated - how easy is it it programmatically determine what version was previously deployed?).

bwhicks:
> @bwhicks thanks for pointing out the GitHub environment concern; I expect reverting to be an emergency measure, but we might not always be able to fix immediately, depending on the problem. If it's not too difficult to update the GitHub deploy, I think that would be valuable (although that seems like it could make it a lot more complicated - how easy is it it programmatically determine what version was previously deployed?).

It should be pretty easy adaptation of the tasks I use to pull in the hash initially anyway. I'll add it to the list of changes.

revert_deploy.yml
thatbudakguy: i feel like i generally run these commands with sufficient anxiety/paranoia such that confirmation prompt isn't super necessary, but fine either way

bwhicks: All right, in that case I've dropped it.

bwhicks:
I believe this has addressed all the concerns, and it is working for me in staging. I had some annoying host key issues with the staging VM when I recreated it (and I'd suggest so doing), but once those were solved locally everything worked as hoped.

PR #11, opened 2019-02-13
Add deploy bash wrapper for QOL improvement

bwhicks:
This PR adds a wrapper command `deploy` that can be used to make deployments to QA/Prod more clearly and easily pull latest changes from `master` and use them.

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,6 +56,24 @@ for secrets in a `local_settings.py`, but prevents worst case scenarios from
 emerging.
 
 
+### Running a deploy
+
+To make running deployments easier, you can use the provided `deploy` script
+to wrap `ansible-playbook`. By default, `deploy` switches you to `master`,
+runs `git pull` and checks for uncommited changes to files, etc. If any appear,
+it aborts.
+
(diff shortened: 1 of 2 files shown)


bwhicks:
@rlskoeser Because I think there's a good chance I'll get to some deploy work during this off week, I thought I'd ask: were you thinking an ansible method or just a bash executable we could get in the habit of running that did these things and asked for a bit of input?

The former would (imo) only be a bit more robust, and the bash script has the benefit of being quite light (and adaptable from previous work).

I'm also happy to icebox this. It was something I dashed off quickly as an idea for how to manage playbook sync.

rlskoeser:
@bwhicks I was thinking a bash script. That seems simpler to me and could be adapted from what you've already written. Sounds like you agree?

I also had a thought of what I would call the two modes, but failed to follow up and comment: check and prep (or something along those lines). So if you keep the bash script called `deploy`, then you would run `deploy check` or `deploy prep`.  I think we could easily get in the habit of that, and it seems like a good improvement over running `git status`.

bwhicks:
Closing out as a not priority

PR #12, opened 2019-02-14
add webpack role

thatbudakguy:
- adds a role for running webpack
- updates playbooks for mep and ppa, incl. staging and qa
- adds media configuration and role to mep
- removes some django-compressor related things that are no longer used

### DIFF
--- a/roles/run_webpack/tasks/main.yml
+++ b/roles/run_webpack/tasks/main.yml
@@ -1,0 +1,29 @@
+###
+# Run webpack to compile static assets (css and javascript).
+#
+# This role runs Webpack to compile and compress source javascript and styles
+# into css/js files. It depends on webpack and all its associated js dependencies
+# having been installed, usually through the 'build_npm' role.
+# 
+# The role expects that appropriate qa/prod-specific tasks will be defined in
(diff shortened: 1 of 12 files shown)


roles/run_webpack/tasks/main.yml
```diff
@@ -0,0 +1,29 @@
...
+#
+# This role runs Webpack to compile and compress source javascript and styles
+# into css/js files. It depends on webpack and all its associated js dependencies
+# having been installed, usually through the 'build_npm' role.
+# 
+# The role expects that appropriate qa/prod-specific tasks will be defined in
+# package.json and callable via `npm run build:qa` and `npm run build:prod`.
+###
+---
+- name: webpack tasks
```
bwhicks: Could this be slightly more descriptive. 'Run webpack build tasks.' Clearly it's provided up top too, but it's often handy to have a bit more descriptiveness in the actual play.

thatbudakguy:
That's a really good question. I think I'm OK adhering to those two tasks as conventions going forward, since behind the scenes I can sort of map them to whatever webpack config we want for any given project (and we'll always have a qa and a prod).

PR #19, opened 2019-10-14
use deploy_environment var for staging checks

### LINKED ISSUE #17: pre- and post-deploy tasks fail when staging var is undefined
checks at the end of `create_deployment` and `close_deployment` are failing when `staging` is not defined:
https://github.com/Princeton-CDH/CDH_ansible/blob/8b971961a366bf9a24a54bd01417e734e3305665/roles/create_deployment/tasks/main.yml#L39

[ansible docs on `when` statement](https://docs.ansible.com/ansible/latest/user_guide/playbooks_conditionals.html#the-when-statement)

see slack link below for more

thatbudakguy:
fixes #17

### DIFF
--- a/roles/close_deployment/tasks/main.yml
+++ b/roles/close_deployment/tasks/main.yml
@@ -14,5 +14,5 @@
           description: "Deployment succeeded"
       headers:
           Authorization: "token {{ github_token }}"
-    when: not staging
+    when: deploy_environment != "staging"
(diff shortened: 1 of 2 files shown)


thatbudakguy:
if you check the link to the slack conversation in #17 you can read through our investigation of this. ultimately it looks like the change I made in this branch was what we settled on.

PR #22, opened 2020-03-03
PEMM QA playbook revisions — apache config, python version

rlskoeser:
- Revised apache config role as discussed (create & enable project-specific conf, disable default conf)
- Revised apache config for working mod wsgi python app with virtualenv
- Configures explicit virtualenv to only run on RedHat; use python version and pip to create with specified python version


rlskoeser:
Updates to refine crontab handling, add logging directory setup for pemm, and new pemm role to copy an encrypted credential file and check out an additional git repository.

PR #26, opened 2020-06-16
Adapt crontab role for springdale playbooks

rlskoeser:


roles/configure_crontab/tasks/main.yml
```diff
@@ -1,23 +1,26 @@
...
+    cron:
+      name: "{{ item.name }}"
+      special_time: "{{ item.special_time }}"
+      job: "{{ item.job }}"
+      user: "{{ deploy_user }}"
+      state: "{{ item.state | default('present') }}"
+    with_items:   # only include items with special time defined
+      "{{ crontab | selectattr('special_time', 'defined') | list }}"
+  # non-special time cron jobs
+  - name: "Configure crontab — (scheduled"
```
thatbudakguy, line 13: looks like a stray "(" in the name.

rlskoeser: oh, yep. they used to be in parenthesis and I was switching to dashes because I thought it would be more readable. thanks

thatbudakguy: agree, it is more readable!

PR #28, opened 2020-07-08
Increase frequency of pemm syncs

kmcelwee:
* I edited `pemm.yml` but I assume it's somehow synced with `pemm_qa.yml`? If one of you could explain this to me I'd appreciate it.
* We might want to consider that if a person is editing the sheet for 20 minutes, 4-5 slack messages will be sent.
* Also, I don't know whether or not we want this to be permanent or a short-term solution.

### DIFF
--- a/group_vars/pemm/vars.yml
+++ b/group_vars/pemm/vars.yml
@@ -62,12 +62,12 @@ app_vars:
 crontab:
 
   - name: "{{ app_name }} synchronize google sheets to git"
-    minute: 1
+    minute: "*/5"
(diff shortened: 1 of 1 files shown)


rlskoeser:
@kmcelwee the configurations in `pemm.yml` are used by both pemm playbooks; configurations specific to qa or production would go into the `pemm_qa` or `pemm_prod` vars. I haven't looked into it enough yet to see if you can override specific parts of a config (like the minutes of a particular cron job) without having declare the whole thing.

That said — I think this should be a very short-term solution, only for active testing, and was thinking it would be short-term enough we wouldn't even want to commit it. (Although perhaps it could be useful as a commented out, documented option...)

kmcelwee:
I'll just scrap it! Thanks for the explanation, and sorry for the confusion

Issue #29, opened 2020-09-15
fix invalid group characters in cdh-web group name (deprecation warning)


rlskoeser:
**To reproduce**
1. Run any ansible playbook
2. See the deprecation warning immediately in the output

> [DEPRECATION WARNING]: The TRANSFORM_INVALID_GROUP_CHARS settings is set to allow bad characters in group names by default, this will change, but still  be user configurable on deprecation. This feature will be removed in version 2.10. Deprecation warnings can be disabled by setting
deprecation_warnings=False in ansible.cfg.
> [WARNING]: Invalid characters were found in group names but not replaced, use -vvvv to see details

**Additional context**
Output from running with `-vvvv` option:
> Not replacing invalid character(s) "{'-'}" in group name (cdh-web_qa)
> Not replacing invalid character(s) "{'-'}" in group name (cdh-web_prod)
> Not replacing invalid character(s) "{'-'}" in group name (cdh-web_prod)
> Not replacing invalid character(s) "{'-'}" in group name (cdh-web_staging)
> Not replacing invalid character(s) "{'-'}" in group name (cdh-web_staging)
> Not replacing invalid character(s) "{'-'}" in group name (cdh-web)
> Not replacing invalid character(s) "{'-'}" in group name (cdh-web)

rlskoeser:
Resolved by pr #31, already merged

PR #31, opened 2020-09-17
Rename cdh-web_* to cdhweb_ to remove invalid character in group names

rlskoeser:
ref #29


thatbudakguy (approved):
lgtm; i didn't see any other places that still had `cdh-web` that ansible was likely to care about.

Issue #30, opened 2020-09-15
Replace get_ver.py script with one-line python command


rlskoeser:
I think we should be able to replace the `get_ver.py` script with a one-line python command that can be run directly from the ansible script.  I think what we want to generate should look roughly like this:
```sh
python -c 'import ppa; print(ppa.__version__)'
```
It will need configuration for the python package name to import and the working directory.

rlskoeser:
this was resolved by PR #32, already merged

PR #32, opened 2020-09-17
Get python/django app version with one-line python shell command instead of external script

rlskoeser:
ref #30


thatbudakguy (approved):
lgtm

PR #33, opened 2020-09-17
use node v10 for cdhweb

thatbudakguy:
- switches cdhweb to use node v10 for QA and production
- enables the `npm install` step (`build_npm` role) for both QA and production deploys

confirmed that node v10 is available in both environments (currently using in QA; installed in prod).

### DIFF
--- a/cdh-web.yml
+++ b/cdh-web.yml
@@ -12,6 +12,7 @@
     - build_project_repo
     - build_virtualenv
     - install_local_settings
+    - build_npm
     - configure_media
     - django_collectstatic
     - django_compressor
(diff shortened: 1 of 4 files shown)


cdh-web.yml
```diff
@@ -12,6 +12,7 @@
     - build_project_repo
     - build_virtualenv
     - install_local_settings
+    - build_npm
```
rlskoeser, line 15: Should this step be added to the cdh-web staging playbook also? (IDK how much we're still using that, but seems best to keep them in sync)

rlskoeser (approved):
Changes look fine to me!

cdh-web.yml
thatbudakguy: couldn't hurt!

PR #34, opened 2020-10-06
Feature/geniza search qa

kmcelwee:
* I've made up hosts, `test-geniza.cdh.princeton.edu` and `geniza.cdh.princeton.edu` but I assume that we'll replace them with something?
* I've called the solr_core `geniza_qa` and have the gitref set to `experiment/search`. I'm not sure how you want to configure deploys from an experimental branch? Or if you wanted to create a deploy branch here for the search prototype?
* I'm starting out on `test-index.cdh.princeton.edu`, that alright?
* I've removed the indexing cron tab since there's no syncing going on. So no crontab needs to be configured? Are we alright having the developer restart an index every time we update the data, at least for now?
* Is the "Ensure bin directory exists." task necessary? If not, `geniza_setup`, without our google sync and cron scripts, is just copying the local settings. Should we replace `geniza_setup` with `install_local_settings`? This would require defining `template_path` and converting `local_settings.cfg` to `local_settings.py`.

While trying to understand ansible, I came across some other questions I'd like to go through when we have time!


rlskoeser:
@kmcelwee sorry I forgot to share the hostnames! They are `cdh-test-geniza1` and `cdh-geniza1 `
Public names will be test-geniza.cdh.princeton.edu and geniza.cdh.princeton.edu.
Other responses:
 - Setting the gitref to our experiment branch sounds like a reasonable solution for now. I haven't decided yet how I want to handle multiple prototypes, IDK if we'll ever need to have more than one running at time.
- Sounds reasonable to use test-index for now, but I'm not sure whether the new qa box will have access to it or not. We can figure out when we get there whether it's easier to get access to that or switch to the new solr.
- Since it's easy, why not set up a cron job to run the indexing? Why do it manually if we don't need to?

I'll have to review your last point more carefully later. I did wonder about whether we needed a geniza setup analogous to the pemm-specific one, and what we might be able to generalize.

group_vars/geniza/vars.yml
```diff
@@ -0,0 +1,47 @@
...
+solr_collection: geniza
+# apache location
+apache_app_path: "/var/www/{{ app_name }}"
+# wsgi file relative to apache location
+wsgi_path: "{{ app_name }}.wsgi"
+# use python 3.6
+python_version: "python3.6"
+# Override clone root to use deploy user home instead of root
+clone_root: "/home/{{ deploy_user }}/repos"
+# for Geniza, do we want to differentiate between qa/prod paths?
```
rlskoeser, line 27: Nope, let's not

group_vars/geniza/vars.yml
```diff
@@ -0,0 +1,47 @@
...
+  - name: GENIZA_SOLR_CORE
+    value: "{{ solr_collection }}"
+  - name: FLASK_APP
+    value: 'scripts/server.py'
+
+# configure scripts to run as cron jobs
+crontab:
+  - name: "{{ app_name }} index metadata"
+    minute: 5
+    job: "bin/cron-wrapper {{ deploy }}/env/bin/flask index {{ deploy }}/data/pgp-metadata.csv >> {{ logging_dir }}/index.log  2>&1"
```
rlskoeser, line 46: Thanks for adding the cron job. Should just be `flask index` once you update to the new version of the indexing script and add local settings.

group_vars/geniza_qa/vars.yml
```diff
@@ -0,0 +1,3 @@
+solr_url: 'https://test-index.cdh.princeton.edu/'
+solr_core: 'geniza_qa'
+gitref: 'experiment/search'
```
rlskoeser, line 3: Let's add a comment here about this. I'm not sure exactly what to put and we're unlikely to leave it for too long, but just as a reminder for ourselves. Maybe something like, "default to flask search app while prototyping" ?

roles/geniza_setup/tasks/main.yml
```diff
@@ -0,0 +1,15 @@
...
+    owner: "{{ deploy_user }}"
+    group: "{{ deploy_user }}"
+
+- name: Ensure bin directory exists.
+  file:
+    path: "/home/{{ deploy_user }}/bin"
+    state: directory
+    mode: 0700
+    owner: "{{ deploy_user }}"
+    group: "{{ deploy_user }}"
```
rlskoeser, line 15: I can't remember why this is here! I think maybe it is a leftover from the cron scripts, which I may have first included here before I generalized for re-use with S&co. I think you're right that we don't need it.  It would be great if we didn't need a separate geniza setup role and could fold the local settings handling into an existing task.

rlskoeser: I just checked — I think you're right, it looks like it should work to switch to using the existing `install_local_settings` role. Let's try that and get rid of this one!

rlskoeser (commented):
Looks good so far!

group_vars/geniza/vars.yml
rlskoeser: (but we can add a comment that we don't, since we do differentiate on the older RC VMs)

group_vars/geniza/vars.yml
```diff
@@ -24,24 +24,19 @@ wsgi_path: "{{ app_name }}.wsgi"
...
 clone_root: "/home/{{ deploy_user }}/repos"
-# for Geniza, do we want to differentiate between qa/prod paths?
+# for Geniza, we don't distinguish between qa/prod paths
 install_root: '/srv/www/{{ app_name }}'
+# path to dir that holds local_settings.py
+template_path: "{{ app_name }}"
+solr_core: 'geniza'
 
 # override default production contexts (no code coverage for this project yet)
 deploy_contexts:
```
rlskoeser, line 34: nice!

group_vars/geniza/vars.yml
```diff
@@ -24,24 +24,19 @@ wsgi_path: "{{ app_name }}.wsgi"
...
     - 'continuous-integration/travis-ci/push'
 
-app_vars:
-  - name: GENIZA_SOLR_URL
-    value: "{{ solr_url }}"
-  - name: GENIZA_SOLR_CORE
-    value: "{{ solr_collection }}"
-  - name: FLASK_APP
-    value: 'scripts/server.py'
-
```
rlskoeser: Do we not need these anymore?

rlskoeser (commented):
Updates look good, just one question about the app vars — are they all handled elsewhere now? I would have thought we'd at least need the flask app setting somewhere, but I forget where this is used.

group_vars/geniza/vars.yml
kmcelwee: Oh, right! I had thought "we don't need environment variables, because everything's in `local_settings.py`" but I had forgotten about `FLASK_APP`. I'll push the fix now.

rlskoeser (approved):

kayiwa (approved):
OK

kmcelwee:
Beginning launch to QA 🚀

---

Error: `Conflict: Commit status checks failed for experiment/search.`  Caused because of `required_contexts` variable: https://developer.github.com/v3/repos/deployments/ Travis setup was ported from PEMM. Replacing with empty array to ignore.

kmcelwee:
Tunnel to view on localhost:8383 with:
```
ssh -L 8383:localhost:80 pulsys@cdh-test-geniza1
```
Currently 404 error

kmcelwee:
- [ ] Double check mep and pemm deploy after https://github.com/Princeton-CDH/CDH_ansible/pull/34/commits/75c6809dd111d24d70626e130406a466e4df482b

PR #39, opened 2020-11-23
PUL harmonization updates

thatbudakguy:
this PR makes a variety of changes inspired by the [PUL ansible repository](https://github.com/pulibrary/princeton_ansible) in preparation for writing more playbooks that operate on PUL infrastructure. comments/questions welcome.

tested by deploying cdh-web to QA with no issues - though part of the reason for opening this PR is to prepare to write more (actual) tests!

- updates to latest ansible
- installs molecule with docker feature for writing automated tests for roles/playbooks
- moves playbooks to the playbooks/ folder
- adds symlinks to the playbooks/ folder so that playbooks run correctly
- adds .python-version file to track preferred python version for local dev and CI
- moves templates for local_settings.py to the `install_local_settings` role and renames them by project
- removes vagrant staging setup since it no longer matches PUL infrastructure (could recreate via docker if we want)


rlskoeser (commented):
Changes all look reasonable to me. I'm not sure how to test anything beyond running a playbook (which you've already done). Do we have a molecule test command / playbook yet that can be added to the readme? 

What's the rationale for moving the playbooks to a subdirectory? Certainly seems tidier. Any idea if this is an ansible convention we weren't following?

Similar question for the local settings change. Is this a convention we weren't following? I don't object to it, just curious. Could we rename the project-specific settings to `appname_settings.py` instead of just `appname.py` ?

I'm completely in favor of removing the vagrant vm; I haven't used it in ages and I don't know how useful it actually ever was. Using molecule will be much more useful. I think this means that _all_ of the staging hosts and playbooks should be removed, along with a few remaining references to staging in the readme. 

I happened to notice a reference to 'master' in the readme, would you mind changing that to main while you're at it?

thatbudakguy:
> Do we have a molecule test command / playbook yet that can be added to the readme?

see #40 for an example role with molecule tests you can run (it depends on this PR). once merged I can add some notes to the README similar to the instructions on #40.

> What's the rationale for moving the playbooks to a subdirectory? Certainly seems tidier.

I think just to not have them all in the root, since there are a lot. Regarding conventions, this is the way PUL does it (including symlinks), but in actuality it seems like storing all your playbooks/roles in a monorepo isn't the convention (although molecule docs do have [some notes about it](https://molecule.readthedocs.io/en/latest/examples.html#monolith-repo) as a common use case). Evidently people treat them more like packages (since they can be shared on ansible galaxy), which is the reason for the namespacing (`cdh.foo`) and keeping them in separate repos/folders.

> Similar question for the local settings change. Is this a convention we weren't following?

Sort of; there's a general convention to associate any templates used by a role with that role in a `templates/` folder. More broadly, I'm still trying to figure out what a role is _for_ - we seem to be treating them like tasks, but PUL treats them almost more like packages or modules, and that seems to be the convention on ansible galaxy as well. It might be that we move to a structure where roles aggregate more functionality and can be used in multiple ways, e.g. we have something like `cdh.django` which does everything necessary to set up a Django app based on some arguments you pass to it.

I'll clean up the README a little more and remove all of the staging stuff. Also will rename the settings files.

rlskoeser (approved):
Thanks for the responses — I approve this to be merged once you make the changes you have planned.

(I've looked at the second PR, but I think it will be easier to review once this is merged)

PR #40, opened 2020-11-24
Example tested role with molecule

thatbudakguy:
<s>branched off of #39, which it requires.</s> rebased against `main` now that #39 is merged.

this PR converts the `build_dependencies` role, which is analogous to PUL's `pulibrary.common`, into `cdh_common`, a role formatted according to modern ansible/molecule conventions and tested with molecule.

this role was scaffolded using the command:
```sh
$ molecule init role cdh_common --driver-name docker
```

several folders generated by molecule can be deleted (e.g. the `tests/` and `handlers/` folders) after this command runs, since few of our roles will use them. the `default` folder is a "scenario", molecule's concept of a test suite - they seem to usually be used to test against different platforms. we will likely only ever use `default`, except perhaps if we want to test roles against both PUL VMs and research computing ones.

for molecule configuration, see comments on `roles/cdh_common/molecule/default/molecule.yml`. this file will probably look identical for all of our roles.

for the actual tests, see `roles/cdh_common/molecule/default/verify.yml`. ansible's docs suggest not bothering to test trivial things such as if a file exists or a service has started, since this is really ansible's core job, but i think this is how to do it (and it doesn't hurt to double-check). it's evidently not a convention to test playbooks, only roles - this seems akin to unit testing vs integration testing.

to run the tests, `cd` to the `roles/cdh_common` folder and then:
```sh
molecule test
```

if you want, you can instead run:
```sh
molecule converge
```
this will create a docker container and execute the role against it, and then stop, letting you get into the container and poke around (usually via `docker exec -it [container] /bin/bash`). `molecule destroy` will clean up when you're done.


playbooks/geniza_qa.yml
```diff
@@ -4,7 +4,7 @@
   become: true
   roles:
     - create_deployment
-    - build_dependencies
+    - cdh.common
```
rlskoeser, line 7: Common is a really generic name, can we please name it something more meaningful/descriptive? That tells me nothing about what it does.

rlskoeser: Should this be `cdh_common` here and throughout?

roles/cdh_common/molecule/default/verify.yml
```diff
@@ -0,0 +1,27 @@
+---
+# This is an example playbook to execute Ansible tests.
+
+- name: Verify
+  hosts: all
+  gather_facts: false
+  vars_files:
+    - ../../defaults/main.yml
```
rlskoeser, line 8: Is there another option besides this relative path? Would prefer it to be explicit and more readable

rlskoeser (commented):
Was able to run the molecule test locally, although I did get an error about lint failing.

Seems annoying that we have to cd into the role directory to run the test, is that temporary?

I also got a non-obvious error at first because docker wasn't running; is that a standard enough thing we don't need to mention it?

thatbudakguy:
closing in favor of #47 which also documents how to do this.

PR #41, opened 2020-11-24
Geniza production deploy playbook

thatbudakguy:
- Set DATA_DIR for geniza qa and prod deploys
- Create geniza prod deploy playbook


rlskoeser (approved):

thatbudakguy:
idk if this is doing anything useful, but we could keep it around until needed?

rlskoeser:
@thatbudakguy isn't this the production playbook for the flask prototype? IDK if it's easier to start with this or work from the new django qa playbook, but I'm ok with keeping it around until we're ready to figure that out. 

Should we do anything with the PR?

thatbudakguy:
oh, good point. i don't think we need it at all then! i'll get rid of it.

PR #44, opened 2020-11-24
Geniza i18n prototype QA deploy playbook

thatbudakguy:
- Extend default django settings
- Remove template_path from geniza vars
- Add group vars for geniza i18n QA
- Add apache_app_url setting to configure_apache role
- Update local_settings role to use default template_path
- Add geniza i18n QA playbook


roles/build_dependencies/tasks/main.yml
```diff
@@ -1,7 +1,7 @@
 ---
 - name: Install configured dependencies
   apt:
-    name: '{{ configured_dependencies }}'
+    name: '{{ common_dependencies + app_dependencies }}'
```
thatbudakguy, line 4: whoa, you can just use `+`? that's pretty handy

roles/configure_apache/defaults/main.yml
```diff
@@ -1,7 +1,15 @@
...
 apache:
-    docroot: '{{apache_app_path |default("/var/www/html")}}'
-    servername: "{{ ansible_hostname }}"
+    docroot: '{{ apache_app_path |default("/var/www/html")}}'
+    servername: '{{ apache_servername | default(ansible_hostname) }}'
     directory_options: ''
-apache_listen_port: 80
-apache_app_url: '/'
\ No newline at end of file
+    listen_port: '{{ apache_port | default("80") }}'
```
thatbudakguy: nice, I like the use of `default()` here so that defaults are actually kept in this file, instead of in the template file.

thatbudakguy (commented):
changes look good, I'll take a crack at adding the required stuff for connecting to mysql.

roles/build_dependencies/tasks/main.yml
rlskoeser: Yeah, that's what I found when I went looking for a solution! I think this is a lot simpler than whatever analogous thing I did before.

roles/configure_apache/defaults/main.yml
rlskoeser: Yeah, seems much better. Do you think it could just be `port` instead of `listen_port` ? Is `port` ambiguous in this context?

thatbudakguy: i can't think of any other meaning in this context, so `port` seems fine to me?

thatbudakguy:
@rlskoeser ready for final review based on changes we made today, if you want

rlskoeser (approved):
lgtm

thatbudakguy:
no longer needed since prototype complete

PR #54, opened 2020-12-15
Additional tweaks for cdhweb qa playbook

rlskoeser:
- sets media root to be owned by apache user + group
- fix permissions error when creating postgresql db dump
- changes db name and password (motivation: db name on postgres server should be recognizable and not so generic)
- adds python image dependencies for mezzanine thumbnails 
- adds bash alias to cd to current deploy and activate python environment


roles/deploy_user/templates/bash_aliases.j2
```diff
@@ -0,0 +1,7 @@
+# bash aliases for {{ deploy_user }} / {{ app_name }}
+# NOTE: this file is {{ ansible_managed }}
+
+# alias to change directory to current deployed version
+# and activate python virtualenv
+alias activate="cd {{ install_root }}/current && source env/bin/activate"
+
```
thatbudakguy, line 7: nice!!

thatbudakguy (approved):
lgtm, love the alias! definitely we should be leveraging more things like that.

rlskoeser:
> lgtm, love the alias! definitely we should be leveraging more things like that.

I've been wanting that alias forever! (the command is even more complicated on Springdale where you have to activate python 3, etc).  I'm so glad we have an easy way to add things like that now.

PR #56, opened 2021-01-05
playbook/cdh web prod pul

thatbudakguy:
- Update cdhweb production playbook for PUL VMs
- Consolidate group vars for cdhweb qa/prod
- Remove README note about unused group_name
- Add python3-dev dependency to postgres role
- Remove staging group vars


thatbudakguy:
something is up with the permissions on the logfile/directory - the deploy is currently erroring because the `django_collectstatic` role is unable to write to the django logfile.

initially, i saw errors in the _apache_ logs about not being able to write to the django logfile. i think i saw that its permissions were `conan`/`users` which seemed weird (what's the `users` group? everyone?) so I changed both its permissions and the log directory permissions to `www-data`/`www-data` to match the permissions in QA. this caused the site to run correctly thereafter, but now the deploys are erroring.

it seems like maybe the apache group might not be getting correctly defined, or otherwise something is happening to incorrectly set (or not set?) the group for some files/dirs such that it's `users`.

thatbudakguy:
tracked this down to the `logs/cdhweb.log` file not being group writable in prod for some reason, which was a discrepancy between prod and qa. it needs to be owned/group `www-data:www-data`, but be group-writable so that `conan` can write to it.

README.md
```diff
@@ -105,7 +105,6 @@ The rough order of creating a playbook is:
   5. Create a YAML directory in `group_vars` that has a name mirroring the new playbook: i.e. if the playbook is `my_playbook_qa`, the name should be the same, with a `vars.yml` and a `vault.yml` (for encrypted variables). You will also want to create a `group_vars` folder for the project to hold variables common to production, qa and staging. It should have the same name as the `project:children` you defined earlier in `hosts`.
   6. Reference any playbook variables and set accordingly. See above under vault variables for how to configure those.
   6. Add roles to the list in the new playbook in the order needed.
-  7. N.B. Make sure you set the `group_name` variable appropriately as some QA only steps are skipped based on `_qa` not being in the name.
```
rlskoeser, line 108: Is this no longer true?

group_vars/cdhweb/vars.yml
```diff
@@ -20,6 +26,12 @@ node_version: "10"
 # django database backend
 db_backend: "postgresql"
 db_host: "{{ postgres_host }}"
-# image libraries needed for mezzanine image handling
+# postgresql database info
+application_dbuser_name: "{{ vault_db_username }}"
+application_dbuser_password: "{{ vault_db_password }}"
+application_db_name: "{{ vault_db_name }}"
+# Database backup location
+db_backup_path: '/home/{{ deploy_user }}/pre-{{ version }}-{{ short_hash }}.sql'
```
rlskoeser, line 34: Nice, better than putting in tmp. Should it go in a subdirectory, maybe?

group_vars/cdhweb_prod/vars.yml
```diff
@@ -1,10 +1,13 @@
...
+email_prefix: "[CDH Web] "
 # allowed hosts for Django
 allowed_hosts:
-  - 'cdh-web.princeton.edu'
-  - 'cdh.princeton.edu'
-  - 'digitalhumanities.princeton.edu'
+  - "cdh-web.princeton.edu"
+  - "cdh.princeton.edu"
+  - "digitalhumanities.princeton.edu"
+  - "cdh-beta.princeton.edu"
```
rlskoeser, line 11: was this one meant to be temporary?

playbooks/cdh-web.yml
```diff
@@ -1,21 +1,19 @@
...
     - build_virtualenv
+    - postgresql
+    - configure_logging
+    - configure_apache
     - install_local_settings
     - build_npm
     - configure_media
     - django_collectstatic
     - django_compressor
-    - backup_database
```
rlskoeser, line 18: Why are we dropping this? Does it not work with postgres?

I would still like to get a snapshot before running migration if we can. (I know we discussed this and I know postgres has a lot more capability than we had before but I don't remember where we ended!)

rlskoeser (approved):
Changes look good, just a couple of questions about things I'm not sure I understanding or remembering decisions we made.

README.md
thatbudakguy: Trying to remember what the deal with this was...I believe I did a global search for `group_name` and found there was no longer any logic that looked at it, so removed this. I'll check again

group_vars/cdhweb_prod/vars.yml
```diff
@@ -1,10 +1,13 @@
...
+gitref: '{{ ref | default("main") }}'
 # email prefix for admin emails
-email_prefix: '[CDH Web] '
+email_prefix: "[CDH Web] "
 # allowed hosts for Django
 allowed_hosts:
-  - 'cdh-web.princeton.edu'
-  - 'cdh.princeton.edu'
-  - 'digitalhumanities.princeton.edu'
+  - "cdh-web.princeton.edu"
```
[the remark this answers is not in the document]
thatbudakguy: you're right, we didn't change it to point to the new VM, so no point in keeping it here

group_vars/cdhweb_prod/vars.yml
thatbudakguy: github isn't doing a good job of showing me which one you mean specifically, sorry

group_vars/cdhweb/vars.yml
thatbudakguy: Probably a good idea, just need to add to the role to try to create the directory it if it doesn't exist

playbooks/cdh-web.yml
thatbudakguy: This might've just been temporarily disabled because of a permissions issue I was having or something, good catch. I'll add it back

README.md
rlskoeser: That's great if we don't need it.

group_vars/cdhweb_prod/vars.yml
rlskoeser: Sorry, I should have just put one comment and been more clear! I meant cdh-beta — was it meant as a temporary name for the migration, and now it will go away (until/unless we need it again)?

What I didn't say, and should have — this seems like a long list of names and it would be nice if we could clean some of them up. I think maybe some of that could happen at the lib-adc/nginx layer, but I'm not sure (and it's probably not urgent!).

thatbudakguy: yes, cdh-beta was meant as temporary. I'll remove it along with the `cdh-web` one

playbooks/cdh-web.yml
thatbudakguy: ah, I keep forgetting! `backup_database` is now a springdale-only role. the `postgresql` role handles backing up with one of its tasks, so no need to include `backup_database` for any PUL-hosted projects.

README.md
thatbudakguy: Confirmed there's nowhere in the codebase that uses `group_name` at all. I don't think it's some kind of ansible builtin/magic, and if we aren't setting it anymore, presume it's OK to remove this too.

group_vars/cdhweb/vars.yml
thatbudakguy: Added some logic to do this to the backup task, will test on next deploy

PR #58, opened 2021-02-17
Geniza django qa

rlskoeser:
- updates the geniza qa playbook to switch from the flask prototype to new django app
- incorporates fixes from geniza i18n pull request (which I propose we do not need to merge, now that we're done with that prototype)
- switches bash aliases to bash profile so that the `activate` alias is reliably available


thatbudakguy (approved):
everything looks good, but i think we need to create a role that runs `python manage.py compilemessages` on each deploy to prepare the translated content (imo really it should be a task inside a `django` role, but that's for later). not sure when the best time to do that is, but maybe it's now?

rlskoeser:
@thatbudakguy I really like the idea of refactoring the roles to make a django role that includes existing django-specific steps and adds a new optional compile messages step. How soon do you think we need this? Could we put it off a bit until you have some time to help think through and plan that refactor?

thatbudakguy:
forgot to respond to this - definitely fine putting it off to think about it more!

PR #61, opened 2021-03-25
add apache redirects for cdhweb qa

rlskoeser:


rlskoeser:
@thatbudakguy do we need redirects for attachment files also?

thatbudakguy (approved):
I don't know if we need redirects for uploaded content – when I was testing it, it looks like wagtail generates the correct URL to serve it to you when you click an attachment, so we could see if it just works.

PR #62, opened 2021-03-25
Tweak cdhweb qa redirect to force ssl; add equivalent redirect for prod

rlskoeser:


thatbudakguy (approved):
changes look good; if we wanted to be fancy maybe we could move the apache_extraconf block up to the top-level cdhweb settings and then use ansible's templating to fill in the changed part (hostname) based on which environment was being used (qa or prod). just an idea

rlskoeser:
> changes look good; if we wanted to be fancy maybe we could move the apache_extraconf block up to the top-level cdhweb settings and then use ansible's templating to fill in the changed part (hostname) based on which environment was being used (qa or prod). just an idea

tbh, felt a little weird to be using 'extraconf' at all, seems like this should just be in the main template. was just using what we'd already set up before. Agree, it would be nicer to have it in one place, easier to manage...

PR #66, opened 2021-04-21
Add passenger role and configure cdhweb & geniza qa playbooks to use it

rlskoeser:
- adapt PUL passenger role for CDH projects + python/django
- configure cdhweb qa and cdhweb playbooks and variables to use passenger+nginx instead of apache+wsgi
- minor updates to existing roles that referenced apache

I've tested the qa playbook and as far as I can tell everything is working the way I want it.

questions:
- PUL passenger role had a dependency on a ruby role, fine to assume we don't need this?
- I manually installed and configured nginx+passenger on the test cdhweb vm before starting the ansible part; how can we be confident the playbook does everything necessary?
- The molecule test doesn't check very much, is there more we can/should be checking for this role?


kayiwa (commented):
OK

roles/passenger/meta/main.yml
```diff
@@ -0,0 +1,18 @@
+---
+galaxy_info:
+  role_name: passenger
+  author: pulibrary
+  description: Install nginx passenger
+  company: Princeton University Library
+
+  license: MIT
```
thatbudakguy, line 8: I put `Apache2` since that's the LICENSE in the root of this repo

roles/passenger/defaults/main.yml
```diff
@@ -0,0 +1,17 @@
...
+nginx_worker_processes: "{{ ansible_processor_vcpus | default(ansible_processor_count) }}"
+nginx_worker_connections: "768"
+nginx_keepalive_timeout: "65"
+nginx_remove_default_vhost: true
+passenger_extra_config: ''
+passenger_enabled: 'on'
+
+# copied from apache role defaults; should these be defined elsewhere?
+media_root: "/var/www/media/"
+font_path: "/var/www/fonts/"
```
thatbudakguy, line 17: i think it makes sense to define them here!

roles/passenger/defaults/main.yml
```diff
@@ -0,0 +1,17 @@
+---
+passenger_server_name: www.example.com
+passenger_app_root: /var/www/app
+passenger_listen_port: '80'
+passenger_startup_file: app/wsgi.py;
```
thatbudakguy, line 5: what's up with the semicolon in this value?

roles/passenger/README.md
```diff
@@ -0,0 +1,57 @@
...
+## Example Playbook
+
+    - hosts: server
+      roles:
+        - { role: passenger }
+
+## License
+
+MIT / BSD
+
```
thatbudakguy, line 54: I took out `LICENSE` here entirely since it's also in the meta `main.yml`. Idk why the role template has so much duplication tbh

roles/passenger/README.md
```diff
@@ -0,0 +1,57 @@
...
+The `passenger_root` for your application (e.g. the `public` folder in a rails app).
+
+    passenger_app_env: production
+
+The app environment passenger will load.
+
+    passenger_root: /usr/lib/ruby/vendor_ruby/phusion_passenger/locations.ini
+    passenger_ruby: /usr/local/bin/ruby
+
+Values for passenger configuration directives inside `nginx.conf`. These defaults should generally work correctly, but if you build `ruby` on its own (as an example), the path to ruby may be different.
```
thatbudakguy, line 32: might be worth updating this to talk about python instead since we won't be building/configuring ruby

roles/passenger/.ansible-lint
```diff
@@ -0,0 +1,15 @@
...
+  - ANSIBLE0002
+  - ANSIBLE0006
+  - ANSIBLE0010
+  - '503'
+  - '305'
+  - '303'
+  - '306'
+  - '301'
+  - '602'
+  - '204'
```
thatbudakguy, line 13: this seems like a lot of stuff...I've disabled one rule so far because it didn't make sense in the context of a test, but we might want to reexamine this at some point and see if we need to be ignoring everything.

thatbudakguy (commented):
Lots of small comments, etc. Main question is what we want this to do: is it a single role for installing and configuring nginx with passenger? Does it always do both? If yes to either, we should maybe change the role name and some of the documentation. If no, maybe we should have an `nginx` role that takes some of the complexity out of this one.

roles/passenger/.ansible-lint
rlskoeser: this was copied from PUL, I haven't looked at these yet actually

important to review before merging? (this gets to your point about avoiding unnecessary tech debt)

roles/passenger/tasks/main.yml
```diff
@@ -0,0 +1,75 @@
+---
+
+- name: install and configure nginx+passenger
+  become: true
+  block:
+      # Variable setup.
+    - name: Phusion | Include OS-specific variables.
+      include_vars: "main.yml"
```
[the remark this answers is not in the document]
rlskoeser: thanks! I noticed that one of the other playbooks we adapted from PUL, will adjust

roles/passenger/tasks/main.yml
```diff
@@ -0,0 +1,75 @@
...
+        update_cache: true
+      changed_when: false
+      # TODO: This changed when should be removed when we upgrade to molecule 3
+
+    # Nginx and passenger installation.
+    - name: Phusion | Install Nginx and Passenger.
+      become: true
+      apt:
+        name: "{{ nginx_passenger_packages }}"
+        state: present
```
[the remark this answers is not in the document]
rlskoeser: passenger is tightly coupled with nginx (at least, that's the deploy option we're using); I don't think it's worth splitting them out right now without a use case for it. the PUL role is named passenger, so I kept it the same. I suppose in theory you could swap out another passenger deploy option for nginx and have it work?

Open to renaming the role to nginx_passenger or similar, but I think passenger is actually the part we care about.

roles/passenger/vars/main.yml
```diff
@@ -0,0 +1,8 @@
+---
+__nginx_user: "www-data"
+nginx_passenger_packages:
+  - nginx-extras
+  - libnginx-mod-http-passenger
+__nginx_passenger_packages_16_04:
```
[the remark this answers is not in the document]
rlskoeser: IDK! I didn't modify these

roles/passenger/.ansible-lint
thatbudakguy: if we can take the time to figure these out i'd like to do it now...looking at a few of them, some are simple formatting-related things (`ANSIBLE0001` is for trailing whitespace), but others like `ANSIBLE0006` are explicit warnings to avoid using particular modules/patterns. that seems like the kind of debt we want to avoid.

the other thing is that specifying them using only the codes like this is also deprecated (i saw a warning about it today); evidently you're supposed to use [rule names from this list](https://ansible-lint.readthedocs.io/en/latest/default_rules.html) instead, which is at least much nicer to read.

could we start by removing all of these and seeing what `molecule lint` fails on, and go from there? then we could at least re-add the rules using their semantic names, if we need to keep them.

.github/workflows/molecule_tests.yml
```diff
@@ -9,6 +9,7 @@ jobs:
         role:
           - deploy_user
           - passenger
+          - python
```
thatbudakguy, line 12: this is already in main, fwiw

thatbudakguy (approved):

roles/passenger/molecule/default/converge.yml
```diff
@@ -0,0 +1,26 @@
...
+    - name: create minimal app dir
+      file:
+        dest: "{{ passenger_app_root }}"
+        state: directory
+        mode: 0655
+    - name: create minimal wsgi file
+      copy:
+        src: "test_wsgi.py"
+        dest: "{{ passenger_app_root }}/{{ passenger_startup_file }}"
+        mode: 0644
```
thatbudakguy, line 21: could these be `prepare.yml` instead of `pre_tasks` here? that seems to be what it's for

roles/passenger/molecule/default/verify.yml
```diff
@@ -0,0 +1,73 @@
...
+        that:
+          - "'syntax is ok' in configtest_result.stderr"
+          - "not configtest_result.failed"
+
+    - name: Test that nginx is running and returns 200
+      uri:
+        url: http://localhost/
+        return_content: yes
+      register: site
+      failed_when: "'Hello World' not in site.content"
```
thatbudakguy, line 73: nice, this is awesome.

.github/workflows/molecule_tests.yml
rlskoeser: I pulled it in from main; not sure why it's getting flagged as new?

roles/passenger/molecule/default/converge.yml
rlskoeser: ah, sure. still learning the conventions!

thatbudakguy: no worries, me too!

PR #68, opened 2021-04-22
Role to manage solr configset and collection

rlskoeser:
solr functionality is working now (configset synced, uploaded to zookeeper, and collection created if it doesn't exist)

no molecule tests yet

some cleanup still needed for variables and the like, and hoping we can come up with a better solution for running the deploy user role on the solr server, but thought it would be useful to create a draft pr for input


roles/deploy_user/files/id_rsa
```diff
@@ -0,0 +1,136 @@
+$ANSIBLE_VAULT;1.1;AES256
+30343436363139383434343533623335353336666463363138363462653966636234303130633330
```
kmcelwee, line 2: Don't we have some automated system that prevents us from committing unencrypted vault files? I can't find it. And if that is the case, would that system know to apply that check to this file?

roles/solr_collection/tasks/main.yml
```diff
@@ -0,0 +1,50 @@
...
+  become: true
+  delegate_to: lib-solr-staging1.princeton.edu  # solr_staging
+
+- name: Update and configure Solr configset
+  block:
+    - name: Copy project Solr configset files for upload to zookeeper
+      # rsync configset from deployed version on remote host to solr server
+      become: true
+      become_user: "{{ deploy_user }}"
+      command: /usr/bin/rsync -a cdh-test-geniza1:{{ deploy }}/solr_conf/ /solr/cdh_solr/cdh_{{ app_name }}_solr_conf
```
kmcelwee, line 19: Presumably this is easier than using ansible's built in rsync command? https://docs.ansible.com/ansible/latest/collections/ansible/posix/synchronize_module.html

roles/solr_collection/tasks/main.yml
```diff
@@ -0,0 +1,50 @@
...
+
+    - name: Get a list of current Solr collections
+      uri:
+        url: "{{ solr_url }}admin/collections?action=LIST"
+      register: solr_collections_list
+
+    - name: Create Solr collection if it does not already exist
+      uri:
+        url: "{{ solr_url }}admin/collections?action=CREATE&name={{ solr_collection }}&collection.configName={{ solr_configset }}&numShards={{ num_shards }}&replicationFactor={{ replication_factor }}&maxShardsPerNode={{ shards_per_node }}"
+      when: solr_collection not in solr_collections_list.json.collections
```
kmcelwee, line 34: I don't see the for-loop where we would iterate through `solr_collections_list`, or where `solr_collection` is defined. Would someone mind explaining this to me?

roles/solr_collection/.travis.yml
```diff
@@ -0,0 +1,29 @@
+---
+language: python
+python: "2.7"
```
kmcelwee, line 3: This isn't the version of python we use, correct? Or was this autogenerated?

roles/solr_collection/README.md
```diff
@@ -0,0 +1,38 @@
+Role Name
+=========
+
+A brief description of the role goes here.
```
kmcelwee, line 4: There are so many autogenerated files in this new setup, we'll probably need to investing some time into deciding how detailed we want to make these files / how we want to structure tests, documentation, and linting going forward

kmcelwee (commented):
Just as a heads up, I didn't really look at the molecule or example files

roles/solr_collection/molecule/default/INSTALL.rst
```diff
@@ -0,0 +1,15 @@
+*******
+Delegated driver installation guide
+*******
```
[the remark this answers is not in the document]
rlskoeser: oh, I didn't even notice that. Do we have a mixture now?

roles/deploy_user/files/id_rsa
rlskoeser: Good catch. There's a precommit hook in `hooks/pre-commit`; instructions for installing it are in the readme, but right now it only looks for vault files by name. It would be good to add this one to the list.

roles/solr_collection/tasks/main.yml
rlskoeser: We couldn't get it to work! I initially tried using it and spent quite a while troubleshooting it when everything I tested manually worked fine, so finally we decided to switch to just using rsync directly.

roles/solr_collection/.travis.yml
rlskoeser: This must have been copied from the example role Francis copied over from PUL! Should be removed since we're not using travis anymore, thanks for flagging

roles/solr_collection/README.md
rlskoeser: I agree — what would be a good way / when would we be at a good point to go through the PUL example together and revise it to match our current best practices and approach?

rlskoeser:
I'm going to merge this and work on the molecule tests in a separate branch.

Issue #63, opened 2021-03-31
As a developer, I want to manage Django tasks through a single role so that I can simplify existing deploy playbooks.


thatbudakguy:
**Is your feature request related to a problem? Please describe.**
Currently, functionality related to deploying Django apps is spread across several roles:
- `configure_logging`
- `configure_media`
- `django_collectstatic`
- `django_compressor`
- `django_migrate`
- `install_local_settings`

**Describe the solution you'd like**
Ideally, these could all be tasks that are part of a single Django-specific role, which could have shared defaults and dependencies on other roles. New Django-related tasks (for example, https://github.com/Princeton-CDH/geniza/issues/117) can also become part of this role.

**Additional context**
- Ansible [`django_manage` plugin](https://docs.ansible.com/ansible/latest/collections/community/general/django_manage_module.html#ansible-collections-community-general-django-manage-module), from `community.general`
- existing [`ansible-role-django`](https://github.com/cchurch/ansible-role-django) on github

rlskoeser:
Should there be a python role? Logging seems like possibly not django-specific (although perhaps our implementation is); what about virtualenv creation?

Will you duplicate existing tasks so that old playbooks are left unchanged but playbooks for applications deployed on PUL VMs would start using the new role? 

Are the linked django ansible plugins and roles useful at all or just examples we can refer to?

thatbudakguy:
> Should there be a python role

Yes! I think we can basically directly convert the `create_virtualenv` role to a python role; it's a dependency of this role. Will do the necessary parts of that in this issue, but eventually the python role should probably also handle installing the relevant version of python.

> Will you duplicate existing tasks

Yep.

> Are the linked django ansible plugins and roles useful at all

The linked plugin is quite useful and we actually already use it in another role; the linked role does too. It just allows running arbitrary `manage.py` commands with a bit of extra checks/helpers. The linked role is more of an example; in my investigation it appeared to be far too generic/featureful and would need a lot of configuration for us to use as-is. I'm referring to it in case it has useful conventions.

rlskoeser:
Great, thanks for elaborating. Sounds good.

Looking forward to having this managed and organized better!

thatbudakguy:
- [ ] write molecule `verify.yml`
- [ ] test against cdhweb in QA

rlskoeser:
implemented in PR #73 ; remaining tasks have been split into separate issues

Issue #48, opened 2020-12-03
allow specifying and installing arbitrary python versions


thatbudakguy:
**Is your feature request related to a problem? Please describe.**
Stuck on python 3.6 because of ubuntu's package manager.

**Describe the solution you'd like**
`build_dependencies` role should enable access to an external apt repo (`deadsnakes` maybe) and then use it to install a version of python specified by `python_version`. this should become the default python.

PR #69, opened 2021-04-23
Create python role (#48)

thatbudakguy:
In writing the django role (#63) it turned out that the only real dependency was python itself; I ended up making a simple role to take care of installing the desired version and create a virtual environment. This is equivalent to our `build_virtualenv` role but for PUL environments, and with a little more functionality.

- Adds a simple role that can install desired python version (including newer versions not in ubuntu repository)
<s>- Role updates symlinks so `python` works correctly in terminal</s>
- Role can create a virtual environment and install dependencies

If we specify this role as a dependency of the `django` role with the following defaults:
```yml
python_app_path: "{{ django_app_path }}"
python_requirements_path: "{{ django_app_path }}/requirements.txt"
```
Then playbooks that use django need only specify `python_version` and do:
```yml
include_role: django
```
And python + the virtual environment + dependencies will be set up automatically, without needing to explicitly include the `python` role.

### DIFF
--- a/roles/python/molecule/default/verify.yml
+++ b/roles/python/molecule/default/verify.yml
@@ -1,0 +1,29 @@
+---
+# Tests for python role.
+
+- name: verify
+  hosts: all
+  gather_facts: false
+  vars_files: ../../defaults/main.yml
+
+  tasks:
+    - name: check python version in virtual environment
+      command: "{{ python_venv_path }}/bin/python --version"
+      register: _python_version
+      changed_when: true
+
+    - name: python version should be 3.8
+      assert:
+        that: _python_version.stdout.startswith("Python 3.8")
+
+    - name: check installed packages in virtual environment
+      pip:
+        name: Django
+        virtualenv: "{{ python_venv_path }}"
+      check_mode: true
+      register: _installed_pkgs
+      changed_when: true
+
+    - name: django should be v3.1 in virtual environment
+      assert:
+        that: "'Django==3.1' in _installed_pkgs.stdout_lines"
--- a/roles/python/tasks/main.yml
+++ b/roles/python/tasks/main.yml
@@ -1,0 +1,5 @@
+---
+# Basic sequence of tasks for setting up a Python app.
+
+- include: install_python.yml
+- include: install_dependencies.yml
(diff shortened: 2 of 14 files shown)


roles/python/tasks/main.yml
```diff
@@ -0,0 +1,5 @@
+---
+# Basic sequence of tasks for setting up a Python app.
+
+- include: install_python.yml
+- include: install_dependencies.yml
```
rlskoeser, line 5: What's the value of splitting these out into separate tasks?  Seems potentially confusing to have to look for different files than main tasks; it isn't like the tasks are very long.

roles/python/molecule/default/verify.yml
```diff
@@ -0,0 +1,29 @@
...
+      stat:
+        path: "{{ python_venv_path }}"
+      register: _venv
+
+    - name: virtual environment should exist
+      assert:
+        that: _venv.stat.isdir
+
+
+
```
rlskoeser, line 29: should the verify check installing python requirements also? probably not too hard to create a requirements file with one or two things (but is it bad to install things just for testing?)

rlskoeser (approved):
Looks good to me, just a couple things I wondered about

roles/python/tasks/main.yml
thatbudakguy: I mentioned this to Francis on our call – it makes a little less sense here than it might otherwise, but I want to follow this structure for most roles that I write. The purpose doesn't make sense in isolation (the role will always just run the tasks in sequence in `main.yml`) but makes sense when you think about roles as collections of functionality. The idea is that you can import the tasks that you need (almost like methods) in the context of a playbook or another role using [`tasks_from`](https://docs.ansible.com/ansible/latest/collections/ansible/builtin/include_role_module.html#parameter-tasks_from), with the intent that if a playbook needed to run only some parts of a role, it'd be easy to do (but impossible if they're all in `main.yml`). A more real-world example would be running django-compressor tasks: that's a separate task that would fall under the `django` role, but doesn't belong in the `main.yml` sequence, and would be called using `tasks_from` by only the playbooks that need it.

roles/python/molecule/default/verify.yml
thatbudakguy: totally agree, and as soon as i started testing this i found out that the tests don't do what I thought they did 😬 so...more tests coming

rlskoeser (approved):

PR #70, opened 2021-04-27
GitHub Actions to run molecule tests

rlskoeser:


.github/workflows/molecule_tests.yml
```diff
@@ -0,0 +1,25 @@
...
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up Python
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.9
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          python -m pip install -r requirements.txt
```
thatbudakguy, line 21: might be worthwhile at some point to add pip caching, since I think it does speed things up a bit. example from cdhweb: https://github.com/Princeton-CDH/cdh-web/blob/main/.github/workflows/unit_tests.yml#L71-L80

.github/workflows/molecule_tests.yml
```diff
@@ -0,0 +1,25 @@
...
+      - name: Test with molecule
+        run: |
+          cd  roles/${{ matrix.role }} && molecule test
```
thatbudakguy, line 24: i think this can just be a oneliner; `run: cd roles/...`

roles/deploy_user/molecule/default/verify.yml
```diff
@@ -1,7 +1,8 @@
 ---
 - name: Verify
   hosts: all
-  check_mode: yes
+  vars:
+    app_name: molecule_test
```
thatbudakguy, line 5: idk what the best practice is for this, but I noticed myself doing this same thing (i.e. duplicating `vars` for tests across `converge.yml`, `verify.yml`, etc). I ended up making a `test.yml` file in the top-level `vars/` directory for the role (which is for "extra" vars) so that i could import it in test playbooks and get rid of the duplication. lmk if that makes sense to you; happy to explain more

thatbudakguy (approved):
looks good! just some suggestions. i like keeping this simple for now and explicitly specifying all the roles to test.

kmcelwee (approved):
exciting stuff 🙂

roles/deploy_user/molecule/default/verify.yml
rlskoeser: yes, I wondered about that. I'll try it

PR #73, opened 2021-05-17
Django role (#63)

thatbudakguy:
this is a minimal implementation that focuses on core tasks for a django install like migrations and installing the local settings file. it follows the same approach i've discussed with francis of breaking functionality up into small tasks, and including the most likely/common sequence of those tasks in `main.yml` so that playbooks are free to create their own task sequence if they choose.

todo:
- [x] write the role's README
- [ ] add a task for running `django-compressor` for usage on cdhweb
- [ ] write tests in `verify.yml` (see comments there for intended test plan)
- [x] run the role against a real QA environment


roles/django/tasks/install_local_settings.yml
```diff
@@ -0,0 +1,10 @@
+---
+# Creates a new `local_settings.py` file based on a template.
+# https://docs.djangoproject.com/en/dev/ref/settings/
+
+- name: install django local settings
+  template:
+    src: "{{ django_local_settings_template }}"
+    dest: "{{ django_app_path }}/{{ django_app }}/local_settings.py"
```
rlskoeser, line 8: This path needs to be configurable for apps that put them in a non-standard location (flask prototypes, projects with split-settings)

roles/django/templates/local_settings.py
```diff
@@ -0,0 +1,95 @@
+# Cryptographic key for signing secrets. Keep the production key hidden!
```
rlskoeser, line 1: Has this changed at all, or just moved to a new location?

roles/django/defaults/main.yml
```diff
@@ -0,0 +1,30 @@
+---
+# main config
+django_app: myapp
+django_app_path: /srv/www/{{ django_app }}
+django_user: "{{ ansible_user }}"
```
rlskoeser, line 5: `ansible_user` might be pulsys rather than conan — is this so you don't have to depend on / be aware of `deploy_user` ? Is your intention that we override it? And if so, can you set our standard as a default in the shared variables file?

rlskoeser (commented):
Looks reasonable. Have you tried running it with a qa app yet? Strongly advise doing that before writing your molecule tests.

Actually, it would be helpful to see the changes needed to switch a playbook over from our old task setup to using this one. It's hard to tell from this how much will change.

roles/django/templates/local_settings.py
thatbudakguy: it did change; i grouped things into sections and did some formatting. there are also more things that are variables in this version vs. the old one.

roles/django/defaults/main.yml
thatbudakguy: yes, yes, and yes 🙂

thatbudakguy:
closing this and moving `django-compressor` tasks to a separate issue.

PR #76, opened 2021-05-20
ppa node fixes

thatbudakguy:
Two changes I needed to successfully deploy PPA to QA. We specified node v10 in the README, but weren't actually using it on VMs. I also got an unpredictable version of `node-sass` that was too new to be compatible with node v10 despite a correct `package-lock.json`, which I determined was because we weren't using `npm ci` to install js dependencies during the deploy.

- Use nodejs v10 for PPA
- Ensure build_npm role installs from package-lock.json


rlskoeser (approved):
ok!

PR #78, opened 2021-05-21
Geniza logging redux

rlskoeser:


roles/install_local_settings/templates/geniza_settings.py
```diff
@@ -12,17 +12,65 @@
...
+    },
+    'loggers': {
+        'django.request': {
+            'handlers': ['mail_admins', 'debug_log'],
+            'level': 'ERROR',
+            'propagate': True,
+        },
+        'geniza': {
+            'handlers': ['debug_log'],
+            'level': 'WARN',
```
kmcelwee, line 66: We've put some of the status info from `import_data` under "INFO". Will this hide that info? Should we raise the level of more important logs?

kmcelwee (approved):
Thank you for doing this 🙂

roles/install_local_settings/templates/geniza_settings.py
rlskoeser: The script output isn't using this logger, it's configuring and using it's own logger based on verbosity level specified when running the script (I think this is a bit confusing, fwiw).

I think warn should be a reasonable default for qa, and we can always temporarily adjust it if we need to debug something.

PR #83, opened 2021-06-17
Configure cron job for nightly solr reindex for geniza qa & prod

rlskoeser:


group_vars/geniza/vars.yml
```diff
@@ -46,3 +46,12 @@ solr_configset: geniza
 # github contexts required for deploy
 deploy_contexts:
   - "Python unit tests"
+
+# configure scripts to run as cron jobs
+crontab:
+
+  - name: "{{ app_name }} reindex"
+    minute: 3
```
thatbudakguy, line 54: just for clarity, is this 12:03 am daily? a comment might be handy to indicate what real time it corresponds to

thatbudakguy (approved):

group_vars/geniza/vars.yml
rlskoeser: yes, good point.

rlskoeser: Actually, I copied from PEMM and just remembered/figured out that it's indexing every hour at 3 minutes past the hour! I'll specify an hour. Also, I checked the logs and discovered that I need to specify `--no-progress` 😆

PR #84, opened 2021-06-22
revisit pre-commit hooks

rlskoeser:
This update switches the local pre-commit hook script out for pre-commit and uses an existing ansible encryption check with our file pattern list. I added documentation about how to install, or update if you had the old commit hook installed.

If you want to test it (either in the current configuration or trying out the commented out hooks), you can use `pre-commit run --all-files`

I looked for other useful pre-commit hooks, and you'll see them commented out — one is failing due to an ansible parsing error (it flagged dummy passwords, which we can mark as ok with a comment, but the script still errors).

There's also an ansible-lint hook, but a) it seems fairly slow and b) it's currently failing on our old revert deploy playbook (which doesn't actually work, maybe what it's flagging is related).


thatbudakguy (approved):
looks good! i followed the instructions to overwrite the old hook and all is well. I don't think we need ansible-lint as a pre-commit hook because `molecule test` runs it, and thus linting will happen in CI for all the roles that we choose to test. as we write more tests, we'll run into the linter errors and fix them.

Issue #86, opened 2021-07-29
Use shallow clone to check out project repo


thatbudakguy:
**Is your feature request related to a problem? Please describe.**
The `build_project_repo` role currently does a full git checkout of the project's repository each time. We don't need the full git history on the target machine; only the most recent state is necessary.

**Describe the solution you'd like**
A shallow clone will allow us to get the files from github without their git history. It should be sufficient to pass [`depth:1` to the `git` module](https://docs.ansible.com/ansible/latest/collections/ansible/builtin/git_module.html#parameter-depth) in ansible.

The relevant lines are here:
https://github.com/Princeton-CDH/cdh-ansible/blob/80891c61640368e843233e213dac2c0184605a0f/roles/build_project_repo/tasks/main.yml#L29-L37

rlskoeser:
resolved in PR #87

PR #87, opened 2021-07-29
Use git shallow branch clone instead of rsync #86

rlskoeser:
I replaced just the rsync step with the git shallow clone, because that's the smallest change that addresses the immediate problem. We could probably replace the full repo checkout (which is what I think you were proposing); I'm just not sure what the implications are and hesitant to make a bigger change for our springdale deploys.

I think Ben's idea with the original setup is that we only have to clone the full repo once, and then once we have it we will only need to pull in incremental changes. I don't actually know if that's any faster than just doing a shallow checkout of the particular version we want.

I had to upgrade ansible to get a version that supported the git options I wanted to use.

Have tested this deploy on both cdh-web qa and derrida qa.


thatbudakguy (approved):
didn't know about `single_branch: true`, good call on that. this looks good.

PR #88, opened 2021-07-30
Add path for git 2.27 to springdale deploys

rlskoeser:
Adding the git 2.27 path for springdale deploys

Because of the way our builds & paths are configured, this currently has to be set in multiple places

~I'm not sure it's actually working correctly; tested it with ppa in qa and still got the warning that the git version is too old for the options we're using.~
I had a path wrong for ppa, and I did not know I needed to also include a new LD_LIBRARY_PATH; I added those and the new deploy worked correctly for PPA in QA

### DIFF
--- a/group_vars/all/vars.yml
+++ b/group_vars/all/vars.yml
@@ -59,11 +59,11 @@ logging_dir: '{{ install_root }}/logs'
 # logging file
 logging_path: '{{ logging_dir }}/{{ django_app }}.log'
 
-# These include the path for rh-python35 and nodejs6
+# These include the path for rh-python35, nodejs6, and git 2.27
 # Paths are picky and spaces for any YAML multiline syntax cause issues
 # These are defaults for most projects and should be overriden as necessary.
 path: '/opt/rh/rh-python35/root/usr/bin:/opt/rh/rh-nodejs6/root/usr/bin:/opt/rh/rh-git227/root/bin/:{{ ansible_env.PATH }}'
-ld_library_path: '/opt/rh/rh-python35/root/usr/lib64:/opt/rh/rh-nodejs6/root/usr/lib64{% if ansible_env.LD_LIBRARY_PATH is defined %}:{{ ansible_env.LD_LIBRARY_PATH }}{% endif %}'
+ld_library_path: '/opt/rh/rh-python35/root/usr/lib64:/opt/rh/rh-nodejs6/root/usr/lib64:/opt/rh/httpd24/root/usr/lib64{% if ansible_env.LD_LIBRARY_PATH is defined %}:{{ ansible_env.LD_LIBRARY_PATH }}{% endif %}'
 python_path: '/opt/rh/rh-nodejs6/root/usr/lib/python2.7/site-packages{% if ansible_env.PYTHON_PATH is defined %}:{{ ansible_env.PYTHONPATH }}{% endif %}'
 
 # currently default for most projects is 3.5
--- a/group_vars/all/vars.yml
+++ b/group_vars/all/vars.yml
@@ -62,7 +62,7 @@ logging_path: '{{ logging_dir }}/{{ django_app }}.log'
 # These include the path for rh-python35 and nodejs6
 # Paths are picky and spaces for any YAML multiline syntax cause issues
 # These are defaults for most projects and should be overriden as necessary.
-path: '/opt/rh/rh-python35/root/usr/bin:/opt/rh/rh-nodejs6/root/usr/bin:{{ ansible_env.PATH }}'
+path: '/opt/rh/rh-python35/root/usr/bin:/opt/rh/rh-nodejs6/root/usr/bin:/opt/rh/rh-git227/root/bin/:{{ ansible_env.PATH }}'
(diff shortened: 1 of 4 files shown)


group_vars/all/vars.yml
```diff
@@ -59,11 +59,11 @@ logging_dir: '{{ install_root }}/logs'
...
 logging_path: '{{ logging_dir }}/{{ django_app }}.log'
 
-# These include the path for rh-python35 and nodejs6
+# These include the path for rh-python35, nodejs6, and git 2.27
 # Paths are picky and spaces for any YAML multiline syntax cause issues
 # These are defaults for most projects and should be overriden as necessary.
-path: '/opt/rh/rh-python35/root/usr/bin:/opt/rh/rh-nodejs6/root/usr/bin:{{ ansible_env.PATH }}'
-ld_library_path: '/opt/rh/rh-python35/root/usr/lib64:/opt/rh/rh-nodejs6/root/usr/lib64{% if ansible_env.LD_LIBRARY_PATH is defined %}:{{ ansible_env.LD_LIBRARY_PATH }}{% endif %}'
+path: '/opt/rh/rh-python35/root/usr/bin:/opt/rh/rh-nodejs6/root/usr/bin:/opt/rh/rh-git227/root/bin/:{{ ansible_env.PATH }}'
+ld_library_path: '/opt/rh/rh-python35/root/usr/lib64:/opt/rh/rh-nodejs6/root/usr/lib64:/opt/rh/httpd24/root/usr/lib64{% if ansible_env.LD_LIBRARY_PATH is defined %}:{{ ansible_env.LD_LIBRARY_PATH }}{% endif %}'
```
thatbudakguy, line 66: we probably shouldn't even have nodejs6 on `PATH`, let alone installed...that's ancient! i don't think we use anything before v8 on any project, and we definitely _shouldn't_ use anything before v10

group_vars/ppa/vars.yml
```diff
@@ -4,8 +4,8 @@
 ---
 # Using python35 and nodejs10 for QA and staging
 # Paths are picky and spaces for any YAML multiline syntax causes issues
-path: '/opt/rh/rh-python35/root/usr/bin:/opt/rh/rh-nodejs10/root/usr/bin:{{ ansible_env.PATH }}'
-ld_library_path: '/opt/rh/rh-python35/root/usr/lib64:/opt/rh/rh-nodejs10/root/usr/lib64{% if ansible_env.LD_LIBRARY_PATH is defined %}:{{ ansible_env.LD_LIBRARY_PATH }}{% endif %}'
+path: '/opt/rh/rh-python35/root/usr/bin:/opt/rh/rh-nodejs10/root/usr/bin:/opt/rh/rh-git227/root/bin:{{ ansible_env.PATH }}'
+ld_library_path: '/opt/rh/rh-python35/root/usr/lib64:/opt/rh/rh-nodejs10/root/usr/lib64:/opt/rh/httpd24/root/usr/lib64{% if ansible_env.LD_LIBRARY_PATH is defined %}:{{ ansible_env.LD_LIBRARY_PATH }}{% endif %}'
```
thatbudakguy, line 8: seems like it would be nice to parameterize these more so that you could individually override python, node, or git versions for apps instead of having to redefine the whole string.

group_vars/mep/vars.yml
```diff
@@ -17,8 +17,8 @@ paris_overlay: 'https://tiles.arcgis.com/tiles/4Ko8f1mCWFLyY4NV/arcgis/rest/serv
 # These include the path for rh-python35 and nodejs8
 # Paths are picky and spaces for any YAML multiline syntax cause issues
 # These are defaults for most projects and should be overriden as necessary.
-path: '/opt/rh/rh-python35/root/usr/bin:/opt/rh/rh-nodejs8/root/usr/bin:{{ ansible_env.PATH }}'
-ld_library_path: '/opt/rh/rh-python35/root/usr/lib64:/opt/rh/rh-nodejs8/root/usr/lib64{% if ansible_env.LD_LIBRARY_PATH is defined %}:{{ ansible_env.LD_LIBRARY_PATH }}{% endif %}'
+path: '/opt/rh/rh-python35/root/usr/bin:/opt/rh/rh-nodejs8/root/usr/bin:/opt/rh/rh-git227/root/bin/:{{ ansible_env.PATH }}'
+ld_library_path: '/opt/rh/rh-python35/root/usr/lib64:/opt/rh/rh-nodejs8/root/usr/lib64:/opt/rh/httpd24/root/usr/lib64{% if ansible_env.LD_LIBRARY_PATH is defined %}:{{ ansible_env.LD_LIBRARY_PATH }}{% endif %}'
```
thatbudakguy, line 21: should we also do MEP prod?

thatbudakguy (approved):

group_vars/all/vars.yml
rlskoeser: I wonder if these are all superseded for specific projects and nothing is actually using the defaults?

group_vars/ppa/vars.yml
rlskoeser: 💯  agree, but these will all go away when we get off RC/Springdal

group_vars/mep/vars.yml
rlskoeser: I don't see any path overrides for mep prod or qa; they're both using this config (since qa actually matches prod for mep, unlike ppa)

PR #90, opened 2021-08-05
Remove winthrop playbooks, vars, settings, & hosts

rlskoeser:
@thatbudakguy want to do a quick check that I'm not removing anything I shouldn't or missing anything that should be removed?


thatbudakguy (approved):
looks good!

PR #94, opened 2021-08-30
new playbook for shxco qa

rlskoeser:


roles/django/templates/local_settings.py
```diff
@@ -20,10 +20,10 @@
...
     "default": {
         "ENGINE": "django.db.backends.{{ django_db_backend }}",
-        "NAME": "{{ django_db_name }}",
-        "USER": "{{ django_db_user }}",
-        "PASSWORD": "{{ django_db_password }}",
-        "HOST": "{{ django_db_host }}",
+        "NAME": "{{ application_db_name }}",
+        "USER": "{{ application_dbuser_name }}",
+        "PASSWORD": "{{ application_dbuser_password }}",
+        "HOST": "{{ application_db_host }}",
```
rlskoeser, line 26: revert change but add mapping in `group_vars/all`

group_vars/shxco_qa/vars.yml
```diff
@@ -0,0 +1,16 @@
...
+allowed_hosts:
+    - "*"
+# email prefix for dev account emails
+email_prefix: '[QA S&co] '
+
+# solr settings
+# TODO: move somewhere common?
+zk_host: "lib-zk-staging1:2181,lib-zk-staging2:2181,lib-zk-staging3:2181/solr8"
+solr_url: "http://lib-solr8-staging.princeton.edu:8983/solr/"
+solr_server: "{{ groups['solr_staging'][0] }}"
```
thatbudakguy, line 16: is it not safe to put these in `/group_vars/all`? they're only for the new solr role, right?

thatbudakguy (approved):

group_vars/shxco_qa/vars.yml
rlskoeser: `zk_host` is only used for the new solr role, but solr url and solr server are used for the old deploys

rlskoeser: was thinking to hold off moving anywhere until we have the migrations done, and then we can clean up a ton of stuff from the old deploys and old variables. (in spite of the TODO comment)

PR #96, opened 2021-09-02
Shxco qa playbook and django role revisions

rlskoeser:
updates to shxco_qa playbook needed for deploy, including:
- include solr_collection role in playbook
- make sure apache fonts are configured correctly
- enable test twitterbot cron job (test twitterbot credentials in qa vault)
- project-specific local settings
- make it possible to customize apache site name and configure so it is not mep

related updates to django role, including:
- default solr config in local settings template
- default logging config and logging path variable for local settings template
- readme


group_vars/all/vars.yml
```diff
@@ -58,6 +58,7 @@ version: '{{ python_app_version.stdout }}'
 logging_dir: '{{ install_root }}/logs'
 # logging file
 logging_path: '{{ logging_dir }}/{{ django_app }}.log'
+django_logging_path: '{{ logging_path }}'
```
thatbudakguy, line 61: is the reason for having two of these so that we can also support RC apps that use `logging_path`?

group_vars/shxco/vars.yml
```diff
@@ -22,13 +22,16 @@ clone_root: "/home/{{ deploy_user }}/repos"
 install_root: "/srv/www/{{ app_name }}"
 # apache location
 apache_app_path: "/var/www/{{ app_name }}"
+# use site name instead of 'mep' for apache site config
+apache_conf_name: "shakespeareandco"
 
+# allow any host, since we're behind PUL load balancer
+django_allowed_hosts:
+  - "*"
```
thatbudakguy, line 30: i think we talked about moving this to the all vars, since only the new django role (and thus no RC apps) will use it

roles/django/README.md
```diff
@@ -1,38 +1,65 @@
...
+^^^^^^^^^^^^
+- `django_email_password`: configuration for **EMAIL_HOST_PASSWORD** in django settings, default `changeme`
+- `django_email_subject`: configuration for **EMAIL_SUBJECT_PREFIX** in django settings, default `[{{ django_app }}]`
+
+csp config
+^^^^^^^^^^
+Configuration for Content Security Policy reporting in django setttings
+
+- `django_csp_reportonly_uri`: endpoint for reporting (QA); default `http://example.com/`
+- `django_csp_enforce_uri`: endpoint for enforcing; default `http://example.com/`
```
thatbudakguy, line 50: this is fantastic; thanks for writing this all up (and sorry you had to do it for a role you didn't create!)

thatbudakguy (approved):
looks great; thanks for adding the documentation to the django role!

group_vars/all/vars.yml
rlskoeser: It's used in the `configure_logging` role, which I'm still using in the new shxco_qa playbook.

Would it make sense to create a logging setup task within the django role ? (although it's not django specific)

group_vars/shxco/vars.yml
rlskoeser: weirdly it seems to be only defined for shxco_qa! do you think we vaulted it for other apps for some reason? 

I'll move it and check/remove settings for the other PUL deploys

rlskoeser: oh, duh — none of the other playbooks are using the django role yet. 

Would it make sense to create issues for some of the things we know about now? like updating playbooks to use the new django role, potentially moving the logging setup, etc.

PR #98, opened 2021-09-08
production playbook for shxco / PUL

rlskoeser:
includes some fixes for shxco QA playbook

- new playbook & variables for shxco production based on combination of shxco qa and mep prod variables
- identified & resolved an apache media problem in the qa site
- add workaround for setuptools python2to3 
- flush handlers to get sshd restart before end of playbook when ssh config is changed


group_vars/all/vars.yml
```diff
@@ -102,4 +102,4 @@ django_db_host: "{{ application_db_host }}"
 
 # configure VMs to use local timezone
 # (requires including pul timezone role in playbook)
-timezone: America/New_York
+timezone: America/New_York
```
thatbudakguy, line 105: editor removed the final newline? that's kinda weird

roles/python/tasks/install_dependencies.yml
```diff
@@ -13,6 +13,12 @@
       - setuptools
     state: latest
 
+- name: Downgrade setuptools if 2to3 is required
+  pip:
+    name: "setuptools<58"
+    virtualenv: "{{ python_venv_path }}"
+  when:  requires_python2to3
+
```
thatbudakguy, line 21: would it not work to just pin setuptools lower in s&co's own `requirements.txt` file, and make sure it comes first?

thatbudakguy (approved):

roles/python/tasks/install_dependencies.yml
rlskoeser: 🤔 I was assuming that setuptools has to be installed first; but I should be able to test this pretty easily.

rlskoeser: just tested, they can't be specified in the same requirements file — the setuptools version needs to be downgraded before installing the package that needs it

group_vars/all/vars.yml
rlskoeser: I must have — I had a variable I added and then removed! I can fix.

PR #108, opened 2021-10-28
296 configure custom driver

kmcelwee:
**Notes**
* This will probably add approximately an hour to the scrape from sleeping alone (1000 points * 1 second + 1000 deep zooms * 2 seconds). 
* I never ran the ansible script, and my ansible skills are iffy, so @rlskoeser I'd appreciate an especially close review of the yaml.
* This contains the encrypted mapfile, which I assume will be handled in a separate PR when we focus on https://github.com/Princeton-CDH/derrida-django/issues/287

https://github.com/Princeton-CDH/derrida-django/issues/296


rlskoeser:
- updated to make the custom driver optional and rename ours to make clear that it is specific to derrida
- moved the iiif map file and renamed to end in `.vault` so our pre-commit check will flag it if we unencrypt and forget

PR #111, opened 2021-11-12
296c configure driver timeout

kmcelwee:
https://github.com/Princeton-CDH/derrida-django/issues/296

Yay, it works! Two changes that are in this PR that is different from my last run of the crawl:
* I've shortened the sleep before and after a hover. I'm guessing that the reason it didn't work before was not because of load time but because of the URL timing out.
* I'm not including the `--behaviorTimeout`. I think it was specific to non-custom driver features? I didn't fully understand the documentation.

In case we need to go back, here is the docker command that was successful:
```sh
# script to kick off browsertrix crawl with yaml config
# docker commands need to be run as root unless otherwise configured
sudo docker run -v /home/pulsys/crawl-config.yaml:/app/crawl-config.yaml \
    -v /home/pulsys/crawls:/crawls/ \
     -v /home/pulsys/custom-driver.js:/app/custom-driver.js \
     webrecorder/browsertrix-crawler crawl \
    --config /app/crawl-config.yaml --rolloverSize 50000000 \
    --useSitemap \
    --workers 2 \
    --behaviorTimeout 0 \
    --timeout 18000 \
    --collection derrida \
    --driver /app/custom-driver.js
```

### DIFF
--- a/group_vars/derrida_crawl_qa/vars.yml
+++ b/group_vars/derrida_crawl_qa/vars.yml
@@ -12,6 +12,10 @@ browsertrix_crawl_repo_branch: staging
 browsertrix_collection: "derrida"
 browsertrix_custom_driver: "derrida-driver.js"
 
+# Set at 5 hours to ensure that the custom driver can finish hovering over all
+#  visualization elements
+browsertrix_timeout_seconds: 18000
+
 # configure urls to be set in the browsertrix crawl config
 browsertrix_crawl_seeds:
    - url: "https://derridas-margins.princeton.edu/"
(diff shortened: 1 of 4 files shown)


group_vars/derrida_crawl_qa/vars.yml
```diff
@@ -12,6 +12,10 @@ browsertrix_crawl_repo_branch: staging
 browsertrix_collection: "derrida"
 browsertrix_custom_driver: "derrida-driver.js"
 
+# Set at 5 hours to ensure that the custom driver can finish hovering over all
+#  visualization elements
+browsertrix_timeout_seconds: 18000
```
rlskoeser, line 17: Great! Please document this as an optional parameter in the readme for the browsertrix role

rlskoeser (approved):
Thanks for your persistence figuring this out. 

Wanted to make sure I understand: because it takes so long to hover over all the data points on the visualization pages, browsertrix was timing out on those pages and only capturing the whatever it finished before the timeout. Is this accurate?

group_vars/derrida_crawl_qa/vars.yml
kmcelwee: ✅ d197c4f5de50bd29c3729a7afa88f104a4134991

PR #114, opened 2021-12-14
Add role to manage geniza purchased fonts

rlskoeser:
included in this pr:
- delete old geniza_setup role that was used for an earlier geniza prototype (no longer in use)
- new role for any custom geniza setup needed, currently only font management
- tasks to create font directory and extract purchased font files, which are vaulted as a zipfile in the ansible repo
- new config option for passenger/nginx to restrict fonts by referrer


kayiwa (approved):
OK

PR #115, opened 2022-01-06
Configure passenger for PGPv3 redirects

kmcelwee:
* Should I be using the `permanent` flag?
* After your solr PR https://github.com/Princeton-CDH/geniza/pull/495 is merged, presumably I should add a line that handles tags as well?


rlskoeser (approved):
This looks good to me.

Answers to your question:
- yes, we want these to be permanent redirects 
- yes, it would be great if you could add the tag redirects now so we don't forget; maybe put a commented out placeholder for the tag list page, since we don't know yet where it will go?

rlskoeser:
@kmcelwee would you rewrite the issue as a user story? (LMK if you want help with phrasing/why). It is actually a user-facing feature and should be testable.

PR #117, opened 2022-04-29
replication playbook 

rlskoeser:
update 2023-01: I've been using this replicate playbook successfully for months and need to get it merged in. I've added minimal documentation to the readme about how to use it as it exists now and preliminary notes on how to setup other projects to work with it (will revise that part as needed the next time I add another project to the playbook). 

----

draft playbook for replicating production data to qa/staging environment; draws on princeton ansible examples for migrating databases

I started developing the playbook for just geniza, although I hope to generalize it eventually. I am developing and testing it syncing from geniza staging to localhost because I didn't want to risk doing anything bad in production  (which was good — I did inadvertently drop my staging db once because I forgot to specify the delegate host!).

I'm running things from the app server rather than the db server, since I also need access to app media files and will need to run migrations. It made more sense to me to get all the remote files and then handle them rather than get db & handle, then get media & handle; but could see the logic in organizing it the other way too.

questions:
- should default localhost/dev settings be checked in? or just provide an example for overriding?
- ....


playbooks/replicate.yml
```diff
@@ -0,0 +1,86 @@
...
+
+    # if we don't clear the database first, loading the dump results in
+    # errors/warnings about constraints that already exist
+    # NOTE: in local setup, remove database fails if django is running (client connected to db)
+    - name: Remove the database "{{ application_db_name }}" on target host {{ dest_host }} if it exists
+      community.postgresql.postgresql_db:
+        name: "{{ application_db_name }}"
+        login_host: '{{ dest_postgres_host | default(omit) }}'
+        login_user: '{{ application_dbuser_name | default(omit) }}'
+        login_password: '{{ application_dbuser_password | default(omit) }}'
```
acozine, line 58: I would be careful about using `default(omit)` here. I'm not sure what would happen if the playbook didn't pass a password for an old database that it tried to delete. The "missing var" error is annoying but it can be useful for debugging when things go wrong.

playbooks/replicate.yml
```diff
@@ -0,0 +1,86 @@
...
+        state: "absent"
+      delegate_to: "{{ dest_host }}"
+
+    - name: Create empty database "{{ application_db_name }}" on target host {{ dest_host }}
+      community.postgresql.postgresql_db:
+        name: "{{ application_db_name }}"
+        login_host: '{{ dest_postgres_host | default(omit) }}'
+        login_user: '{{ application_dbuser_name | default(omit) }}'
+        login_password: '{{ application_dbuser_password | default(omit) }}'
```
acozine, line 67: I would be careful about `default(omit)` here too. What is the behavior you want if the playbook doesn't pass a password for the new db? Do you want a passwordless db to get created?

playbooks/replicate.yml
```diff
@@ -0,0 +1,86 @@
...
+        owner: '{{ application_dbuser_name }}'
+      delegate_to: "{{ dest_host }}"
+
+    - name: Load the database "{{ application_db_name }}" on target host {{ dest_host }}
+      community.postgresql.postgresql_db:
+        name: "{{ application_db_name }}"
+        encoding: 'UTF-8'
+        login_host: '{{ dest_postgres_host | default(omit) }}'
+        login_user: '{{ application_dbuser_name | default(omit) }}'
+        login_password: '{{ application_dbuser_password | default(omit) }}'
```
acozine, line 78: Same here.

acozine (commented):
Overall looks good - I like all the comments. I put in a couple of suggestions to make it even better.

playbooks/replicate.yml
```diff
@@ -0,0 +1,86 @@
+- hosts: geniza_qa
+  connection: ssh
+  remote_user: pulsys
+  # override default location for db backup
+  vars:
+    - timestamp: "{{ ansible_date_time.date }}"
```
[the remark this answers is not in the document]
rlskoeser: I do use it! But I only noticed when I went looking that I *stashed* the tasks and variables for syncing media files instead of committing what I had, because it was still very rough.

playbooks/replicate.yml
```diff
@@ -0,0 +1,86 @@
...
+
+    - name: Make sure unzipped database dump does not exist (if it does, bunzip2 will fail)
+      ansible.builtin.file:
+        path: "{{ dest_backup_path_unzipped }}"
+        state: absent
+      delegate_to: "{{ dest_host }}"
+
+    # postgres module can create zipped dump file, but at least on OSX
+    # it fails to unzip when attempting to restore;
+    # use command line since builtin unarchive task does not handle bz2
```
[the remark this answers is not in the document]
rlskoeser: Zip has the same problem (unarchive task does not handle it because it is not technically an archive like a tar file is). I went with bz2 because the compression is better and files are smaller than zip. Should I clarify in the comment?

playbooks/replicate.yml
rlskoeser: ah, thanks. I guess I copied/adapted from your playbook without really thinking about it. Will remove them all.

playbooks/replicate.yml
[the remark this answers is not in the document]
acozine: Ah, in that case I would probably just use gzip/tar. If you prefer bzip, a comment like "we use bzip for Reason X, using bzip means we have to use the command module to unzip" would preserve that thinking for Future Us.

playbooks/replicate.yml
```diff
@@ -0,0 +1,128 @@
...
+
+    - name: Extract media to destination media directory
+      ansible.builtin.unarchive:
+        src: "{{ dest_media_backup_path }}"
+        dest: "{{ dest_media_dir }}"
+      delegate_to: "{{ dest_host }}"
+
+
+# could django migration be done as a second playbook? use existing role/task
+- hosts: localhost    # local won't actually work without some setup...
```
acozine, line 119: If you always want to run the migration tasks when you run the db backup/restore, you can probably use `import_tasks` here. 

`- import_tasks: ../../roles/django/tasks/migrate.yml`

See https://docs.ansible.com/ansible/latest/user_guide/playbooks_reuse.html#re-using-files-and-roles for more details.

playbooks/replicate.yml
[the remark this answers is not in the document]
rlskoeser: I messed with this a bit more and I think I remember more what the problem is — the postgres action can create a gzipped file (and also a tar file according to the docs); but the unarchive command doesn't seem able to handle the archive created by the postgres command. 

It does seem nicer to use archive/unarchive, though. Would it be better to just create a `.sql` file with the postgres action and then use `archive` and `unarchive?

rlskoeser:
Revised with a new approach: replicate playbook has separate plays that run on the source machine (generate the sql & media dumps) and the target machine (retrieve and process the generated files). I had a version of this working to update my local dev; I have the sync to qa nearly working, and I'm starting to see how it could generalize so you could run it for multiple apps and use the limit to specify the host group to sync.

quadrismegistus (approved):
Looks great! Very intuitive readme notes and a powerful and helpful playbook.

PR #124, opened 2022-11-10
Correctly set permissions for PPA data files

rlskoeser:


quadrismegistus (approved):
oof— tough little permissions bug. nice, LGTM

roles/prosody_setup/tasks/main.yml
```diff
@@ -1,14 +1,15 @@
...
-- name: Ensure HathiTrust pairtree directory exists.
+- name: Ensure HathiTrust pairtree directory exists and has correct permissions
   tags: setup
   ansible.builtin.file:
     path: "{{ hathitrust_pairtree_path }}"
     state: directory
-    mode: 0744
-    owner: "{{ deploy_user }}"
-    group: "{{ deploy_user }}"
+    mode: 0775
```
quadrismegistus, line 9: So this was crucial change? Giving write and execute status to group?

quadrismegistus (approved):

roles/prosody_setup/tasks/main.yml
rlskoeser: I think the more important change was moving it out from under `/home/conan` — you can't read a directory if you can't execute the directories that it belongs to, and I wasn't setting the home directory permissions anywhere in prosody setup. This is complicated by the fact that home directories have other constraints on permissions, like ssh security won't allow you to use a key if your permissions are too open. Finally figured out it made more sense to move it elsewhere (which I think is actually closer to the location where it was before, not sure why I tried to change it).

PR #126, opened 2022-11-18
Add geniza metadata export settings and cron job for qa

rlskoeser:


roles/django/templates/geniza_settings.py
```diff
@@ -21,9 +21,14 @@
...
 ANNOTATION_BACKUP_GITREPO = "{{ annotation_backup_gitrepo }}"
-# local path where git repo should be cloned
+# local path where annotation git repo should be cloned
 ANNOTATION_BACKUP_PATH = "{{ annotation_backup_path }}"
 
+# git repository for metadata backup/export
+METADATA_BACKUP_GITREPO = "{{ metadata_backup_gitrepo }}"
+# local path where metadata git repo should be cloned
+METADATA_BACKUP_PATH = "{{ metadata_backup_path }}"
+
```
quadrismegistus, line 31: Nice. Cool that ansible can plug in so nicely with django in terms of setting initial settings.

group_vars/geniza/vars.yml
```diff
@@ -128,3 +132,7 @@ crontab:
     minute: "*/5"   # run every five minutes
     job: "env DJANGO_ENV=staging bin/cron-wrapper {{ passenger_app_root }}/env/bin/python{{ python_version }} {{ passenger_app_root }}/manage.py sync_annotation_export -v 0 >> {{ logging_dir }}/sync_export.log 2>&1"
     state: present
+  - name: "{{ app_name }} metadata export "
+    minute: "*/30"   # run every thirty minutes for now (footnote export takes ~10 min)
+    job: "env DJANGO_ENV=staging bin/cron-wrapper {{ passenger_app_root }}/env/bin/python{{ python_version }} {{ passenger_app_root }}/manage.py export_metadata -ws -v 0 >> {{ logging_dir }}/sync_metadata_export.log 2>&1"
+    state: present
```
quadrismegistus, line 138: This is pretty frequent! Are we worried that if it's any longer our commit messages would be less informative?

quadrismegistus (approved):

group_vars/geniza/vars.yml
rlskoeser: I was thinking more granular would be better, esp. since it shouldn't do anything when nothing has changed, and would make it a little easier to see who is doing what. What frequency were you thinking?

rlskoeser: Also, more frequent makes it easier to test. I'm going to go ahead and merge, but I want to hear what you think about frequency, we can always scale back when we turn this on in production.

PR #129, opened 2022-11-29
Changing prosody python version to 3.9

quadrismegistus:


quadrismegistus:
Seems to be working with just this change. 

Note: if there's already a deploy with the same commit hash that environment must be deleted in order for the new virtualenv (and python version) to be installed.

quadrismegistus:
Update: Turns the python version is correctly updating in the django directory and virtualenv but we're now breaking the apache-django connection. Looking into it via issue above.

quadrismegistus:
Final update: @rlskoeser made this same fix in a related PR (https://github.com/Princeton-CDH/cdh-ansible/pull/131), meant to solve the apache error (by switching us over to nginx, a preexisting priority). Closing this PR as redundant

PR #131, opened 2022-11-29
Switch prosody deploy from apache2 to nginx/passenger

rlskoeser:
per https://github.com/Princeton-CDH/ppa-django/issues/501

in this pr:
- cherry-picked the python version change from the other branch
- revise prosody playbooks to remove apache role and add passenger
- add passenger config variables for prosody and prosody qa
- add handler to passenger to ensure that apache2 is not running when we restart nginx

### DIFF
--- a/group_vars/prosody/vars.yml
+++ b/group_vars/prosody/vars.yml
@@ -28,7 +28,7 @@ apache_app_path: "/var/www/{{ app_name }}"
 
 # set passenger defaults for production; override for other environments
 passenger_app_root: "/var/www/{{ app_name }}"
-passenger_server_name: "geniza.princeton.edu"
+passenger_server_name: "prosody.princeton.edu"
 passenger_startup_file: "{{ app_name }}/wsgi.py"
 passenger_python: "{{ passenger_app_root }}/env/bin/python"
 
--- a/group_vars/prosody/vars.yml
+++ b/group_vars/prosody/vars.yml
@@ -11,8 +11,8 @@ django_app: ppa
 symlink: ppa
 # wsgi file relative to deploy location
 wsgi_path: "{{ django_app }}/wsgi.py"
-# use python 3.6
-python_version: 3.6
+# use python 3.9
+python_version: 3.9
 # nodejs version
 node_version: "10"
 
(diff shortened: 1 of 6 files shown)


group_vars/prosody/vars.yml
```diff
@@ -26,6 +26,13 @@ install_root: "/srv/www/{{ app_name }}"
 # apache location
 apache_app_path: "/var/www/{{ app_name }}"
 
+# set passenger defaults for production; override for other environments
+passenger_app_root: "/var/www/{{ app_name }}"
+passenger_server_name: "geniza.princeton.edu"
```
quadrismegistus, line 31: geniza?

group_vars/prosody_qa/vars.yml
```diff
@@ -4,7 +4,9 @@
 ---
 
 # email prefix for dev account emails
-email_prefix: '[PPA S&co] '
+email_prefix: '[PPA QA] '
+
+passenger_server_name: "test-ppa.cdh.princeton.edu"
```
quadrismegistus, line 9: where are these passenger server names coming from?

roles/passenger/handlers/main.yml
```diff
@@ -1,7 +1,16 @@
 ---
+# to automate transition from apache to nginx:
+# handler to stop and disable apache if the service is defined
+- name: Stop apache2
+  ansible.builtin.service:
+    name: apache2
+    enabled: false
+    state: stopped
```
quadrismegistus, line 8: Nice!

quadrismegistus (approved):

group_vars/prosody/vars.yml
rlskoeser: 🤦‍♀️ copy pasto

rlskoeser: (thanks for catching!)

group_vars/prosody_qa/vars.yml
rlskoeser: not sure I understand what you mean? it goes in the passenger config to identify the server

possibly this is redundant and set elsewhere that we could reuse, though

quadrismegistus: I see. So we're using a separate server entirely for passenger, i.e. separate from cdh-test-prosody1 or equivalent ?

rlskoeser: oh, I understand the question now! Yes, public names / aliases are different from the servers where the code is actually running.

PR #133, opened 2023-01-09
Rekey all vaulted files with new vault password (thanks, lastpass)

rlskoeser:
- generated a new vault password and updated the secure note in LastPass
- rekeyed all vaulted files with the new vault password (thanks to PUL for figuring out the syntax: `ansible-vault rekey --ask-vault-password $(grep -Frl '$ANSIBLE_VAULT;')`)

Please confirm you can decrypt vaulted files with the new password (no need to check all, just checking a few should be sufficient)


quadrismegistus:
No, I did not update my password. And woops yeah I mispelt `revault`.

rlskoeser:
ok, second try!

I've updated to follow PUL approach, using lastpass cli and lastpass-ansible. Their setup instructions are here:

https://github.com/pulibrary/princeton_ansible#automatically-pull-vault-password-from-lastpass

We probably should add some of this documentation to our readme, but I'm not sure how much. Input appreciated!

The new vault password is in lastpass, so if you follow the install instructions at the link above, then you ought to be able to do `ansible-vault edit ...` on one of our vaulted files to confirm you can decrypt; to double-check, you can also do the same thing with `ansible-vault edit --ask-vault-password` and enter the password manually from lastpass.

PR #135, opened 2023-01-10
Update vaulted files with new vault password

rlskoeser:
trying again to rekey vaulted files with new ansible vault password

Now using lastpass cli based on what PUL is doing; refer to their setup instructions here: https://github.com/pulibrary/princeton_ansible#automatically-pull-vault-password-from-lastpass

I haven't updated the docs yet and we should, but I'm going to modify this approach slightly when I split out the geniza credentials — I think we won't need the lastpass-ansible gem. LMK what you think needs to be in our local documentation

FYI: AFAICT, when you tell ansible to prompt for a password but have a default password file configured, if you give it a bad password but the default password file is valid, it will decrypt. I had to remove my env var for the password file to be able to test properly that I could decrypt things with the new key and not the old one. (It was decrypting unexpectedly when I entered the old key at the prompt, I presume because it was falling back to the default file.)


quadrismegistus:
Confirmed that old password fails and new password works.

PR #141, opened 2023-06-23
Configure geniza qa to use solr9 staging

rlskoeser:


rlskoeser:
updated the shard steps and made them conditional on solr version being lower than 9; tested on geniza qa playbook and ppa qa playbook that it works for both solr 9 and solr 8

PR #142, opened 2023-08-02
Update test-prosody and prosody for jammy jellyfish and NFS for media/data

rlskoeser:
changes so far:
- add cdh-test-prosody2 to prosody_qa hosts definition
- remove python-virtualenv from default ubuntu packages list (not installable); added python3 packages, and switched the one python call outside of a python virtualenv to use `python3` instead of `python`
- make stop apache2 handler conditional on the service actually being installed

remaining problems:
- I had to reload nginx manually to get the new application to respond; this is not a new problem but thought I would mention (I made a change a while back to try to set it so the handler always runs but it seems like I must have missed something)
- it appears that the new default shell for `pulsys` is `zsh` instead of `bash`; this is ok, but I would like some help updating my existing bash profile settings and alias to work with zsh. (Is it fine for conan to still be a bash user?)

question: is there a better way to manage hosts so we can run only on the new host? or is it best to run playbook on both?


rlskoeser:
Ran into an error with the stop apache 2 error - let's check if there are any apps that still use apache, and remove this if we can. We don't expect to have apache2 installed on our VMs anymore, so this is unnecessary.

group_vars/all/vars.yml
```diff
@@ -125,3 +125,5 @@ db_backup_filename: "{{ timestamp }}_{{ application_db_name }}.sql.bz2"
 db_backup_path: "{{ backup_dir }}{{ db_backup_filename }}"
 media_backup_filename: "{{ timestamp }}_{{ app_name }}_media.tar.gz"
 media_backup_path: "{{ backup_dir }}{{ media_backup_filename}}"
+## nfs
+nfs_server_staging: "128.112.203.82"
```
rlskoeser, line 129: @kayiwa do you want to define this in `hosts` and pull it in from there? or does it need to be referenced by IP?

rlskoeser: another thought: we could define this in `group_vars/qa/vars.yml` as just `nfs_server` and use the same variable with different values for staging/prod hosts

roles/prosody_setup/tasks/main.yml
```diff
@@ -29,3 +29,31 @@
...
+- name: Ensure the presence of cdh qa mount
+  tags: setup
+  ansible.posix.mount:
+    src: "{{ nfs_server_staging }}:/var/nfs/cdh"
+    path: "/mnt/nfs/cdh"
+    state: mounted
+    fstype: nfs
+    opts: rw,sync,hard
+  when:
+   - "'test' in inventory_hostname"
```
rlskoeser, line 48: We usually test based on the `qa` variable, which is set in qa vars for all hosts in the qa host group.  (Although we could revisit that; seems like it would be cleaner to have a variable that was set to something like staging/production)

```suggestion
   - qa is defined
```

rlskoeser: alternately: if we go with my proposal of defining one `nfs_server` variable with different values for qa and prod, the conditional goes away here and we just mount the appropriate server for the environment

rlskoeser (commented):
@kayiwa thanks for the changes for nfs. I have a couple of comments. Let's figure out next steps and how to test!

roles/prosody_setup/tasks/main.yml
```diff
@@ -33,3 +33,41 @@
...
     group: "{{ webapp_user }}"
     recurse: true
+
+- name: Ensure the presence of nfs-common
+  tags:
+    - setup
+    - nfs
+    - never
+  ansible.builtin.apt:
+    name: nfs-common
```
rlskoeser, line 43: this should be added to the list of default packages in the `build_dependencies` role

kayiwa (approved):
OK! Thanks @rlskoeser

PR #145, opened 2023-09-01
Playbooks for sandbox & geotaste

quadrismegistus:
This PR adds a playbook for geotaste, which is our first playbook deploying to the sandbox server; future apps will want to refactor this a bit.


hosts
```diff
@@ -36,6 +36,9 @@ cdh-web1.princeton.edu
 cdhweb_qa
 cdhweb_prod
 
+[geotaste]
+cdh-dev-sandbox1.princeton.edu
```
rlskoeser, line 40: Noting for myself that we still have to figure out how to split out host from app; we may need to add app-specific group vars. I think it's fine if we tackle this when we write a second playbook for the sandbox (probably / hopefully for simrisk)

roles/finalize_deploy/handlers/main.yml
```diff
@@ -10,12 +10,12 @@
...
-  listen: "restart web server"
+# # Restart apache on Ubuntu systems when apache role is in use
+# - name: restart apache
+#   become: true
+#   service:
+#     name: apache2
+#     enabled: true
+#     state: restarted
+#   when:  ansible_distribution == 'Ubuntu' and apache_app_path is defined
+#   listen: "restart web server"
```
rlskoeser, line 21: I'd like to get rid of the apache2 bits as part of the VM upgrade that Francis is helping us with; I'm not sure we can merge this change into `main` until we resolve that, though.

roles/passenger/templates/geotaste_proxy.conf.j2
```diff
@@ -0,0 +1,12 @@
+server {
```
rlskoeser, line 1: is this proxy template no longer needed?

roles/passenger/templates/geotaste_passenger.conf.j2
```diff
@@ -0,0 +1,106 @@
...
+  gzip_static on;
+
+  gzip_proxied        expired no-cache no-store private auth;
+  gzip_disable        "MSIE [1-6]\.";
+  gzip_vary           on;
+
+  {% block root %}
+  location /sco/1 {
+    auth_basic "Restricted Content";
+    auth_basic_user_file /etc/nginx/.htpasswd;
```
rlskoeser, line 55: If we were doing auth, we'd want the password file to be stored in ansible vault and put on the server as part of the playbook.  I'd also want to figure out a way that it would only prompt for the password when you aren't on vpn, but that doesn't matter now.

let's remove these lines.

roles/geotaste_setup/tasks/main.yml
```diff
@@ -0,0 +1,16 @@
+---
```
rlskoeser, line 1: it doesn't look like you have any app-specific setup needed? so this role can be removed

roles/passenger/templates/geotaste_passenger.conf.j2
```diff
@@ -0,0 +1,106 @@
+server {
```
rlskoeser, line 1: what did you have to configure from the base template? it's hard to tell what you changed. Do you still need a custom passenger conf if you aren't doing the proxying?

rlskoeser (changes requested):

quadrismegistus:
Iceboxing the sandbox 🤣 for geotaste for now. Its nginx and related configs will be continued into/superceded by https://github.com/Princeton-CDH/cdh-ansible/pull/149

PR #149, opened 2023-09-25
Simulating risk ansible deployment

quadrismegistus:


group_vars/simulatingrisk/vars.yml
```diff
@@ -0,0 +1,48 @@
...
+# use python 3.9
+python_version: "3.9"
+# Override clone root to use deploy user home instead of root
+clone_root: "/home/{{ deploy_user }}/repos"
+# for simulatingrisk, don't differentiate qa/prod paths
+install_root: '/srv/www/{{ app_name }}'
+
+# override default production contexts (no code coverage for this project yet)
+deploy_contexts:
+    # - 'continuous-integration/travis-ci/push'
```
rlskoeser, line 26: If context checks are failing, this variable may be set wrong somehow; what I expect is that this should be getting QA variables, which disables the context checking before deploys -this line may be overriding that, but it seems like it should be setting it to nothing.

You can run ansible in debug mode to get a lot more detail about what it's doing, that may help figure out whats' going on.

rlskoeser:
reworked and implemented in #152

PR #150, opened 2023-10-09
Update Geniza production playbook to use solr9 production

rlskoeser:


kayiwa (approved):
OK! Thanks @rlskoeser

PR #151, opened 2023-10-11
Add an optional pause before finalizing deploy

rlskoeser:
this change adds an optional pause just before the newly deployed version is made live; must be enabled by specifying with tags

I successfully used this code to deploy an update to the geniza project that switches it to the solr 9 instance, and used the pause to index content into the new solr collection so that there was no downtime with an empty search on the production site while we indexed after the deploy went live.

### DIFF
--- a/roles/finalize_deploy/tasks/main.yml
+++ b/roles/finalize_deploy/tasks/main.yml
@@ -9,6 +9,30 @@
   become: true
   become_user: "{{ deploy_user }}"
   block:
+    - name: Create "next" symlink for any final steps before switching live
+      file:
+        src: "{{ deploy }}"
+        dest: "{{ install_root }}/next"
+        state: link
+      tags:
+        - never
+        - final-pause
+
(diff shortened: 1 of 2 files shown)


roles/finalize_deploy/tasks/main.yml
```diff
@@ -9,6 +9,30 @@
...
+      tags:
+        - never
+        - final-pause
+
+    - name: pause before switching live
+      ansible.builtin.pause:
+        prompt: "Do any steps needed before the new version is live, e.g., reindexing. (CTRL-C when done)"
+      tags:
+        - never
+        - final-pause
```
acozine, line 26: I like using `never` with a second tag - that's a nice way to make the playbook run or not run those tasks in the right scenarios and also have the tags you pass feel more sensible.

acozine (approved):
This is an intriguing solution for smooth deploys. I haven't tested it, but it looks good!

PR #152, opened 2023-10-11
playbook to install simrisk solara app on sandbox vm

rlskoeser:
playbook to deploy simulating risk app on the sandbox vm

- adds and uses an external supervisor role for configuring supervisor to manage the simrisk solara app
- adds a new sandbox role; adds a default root html site with splash page so multiple apps can live at other urls underneath
  - modifies passenger/nginx role to allow skipping the app-specific passenger site config, so that sandbox nginx is compatible with existing passenger/nginx setup used by other cdh python apps
- modifies deploy user role to accommodate new situation where one host has multiple apps; configure multiple aliases for activating app-specific virtual env in current deploy
- playbook and variables for deploying simulating risk to sandbox with supervisor
- generalizes a few settings that were being overridden everywhere to standard values or variables that were named too specifically (python vs django)


kayiwa (approved):
OK! thanks @rlskoeser

PR #153, opened 2023-10-13
Simplify how setup tasks are skipped

rlskoeser:
use `never` to skip setup tasks by default, rather than configuring it as a tag to be skipped by default (which is harder to override)

I haven't run the playbooks with this configuration, but I used the `--list-tasks` option to check and it looks like it's configured properly

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,16 +77,13 @@ The playbook will run, noting success and failures. The `-v` flag adjusts verbos
 
 ### Skip setup tasks
 
-By default, initial provisioning and setup tasks are configured to be skipped.  Tasks or groups of tasks should be tagged as `setup`.
+By default, initial provisioning and setup tasks are configured to be skipped.  Tasks or groups of tasks should be tagged with both `setup` and `never`.
 
 To run playbook without skipping setup tasks, override the default skip tag configuration:
 
 ```{bash}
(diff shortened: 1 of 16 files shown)


roles/deploy_user/tasks/main.yml
```diff
@@ -4,7 +4,8 @@
   become: true
   tags:
     - setup
-    - cdhsetup
+    - cdhsetup  # ?
```
acozine, line 7: What does this tag do?

acozine (commented):
Overall looks great!

README.md
```diff
@@ -77,17 +77,14 @@ The playbook will run, noting success and failures. The `-v` flag adjusts verbos
 
 ### Skip setup tasks
 
-By default, initial provisioning and setup tasks are configured to be skipped.  Tasks or groups of tasks should be tagged as `setup`.
+By default, initial provisioning and setup tasks are configured to be skipped.  Tasks or groups of tasks should be tagged with both `setup` and `never`.
 
 To run playbook without skipping setup tasks, override the default skip tag configuration:
```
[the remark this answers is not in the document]
rlskoeser: thank you for catching this!

roles/deploy_user/tasks/main.yml
rlskoeser: I don't think it does anything! I just searched through all our roles and I don't see it used anywhere else. I'll remove

PR #159, opened 2023-10-24
New playbook to run database setup and update django local settings

rlskoeser:
- update postgres host variables for new servers
- update postgres version to 15
- new playbook to run postgres database setup tasks and update django local settings

## notes:
- specific tags are required for the new playbook to operate the way it's intended; I have it documented in the comments, but it bothers me a little
- I'm understanding ansible better, and I know now from trying it that this playbook doesn't work properly when run on multiple hosts at a time. I suspect this is because we have tasks using `register` where we should be using `set_fact`; this is something I'd be interested in improving


roles/python/tasks/install_dependencies.yml
```diff
@@ -42,7 +42,9 @@
   register: _pyproject_file
 
 - name: Install python dependencies from pyproject
-  when: _pyproject_file.stat.exists
+  when:
+    - not _requirements_file.stat.exists
```
rlskoeser, line 46: this is an unrelated change: recently added logic to install python dependencies from a different config file (needed for simulating risk), but when running this discovered that an additional check is needed for when to run that install (geniza has the pyproject file to configure a tool but doesn't use it for python requirements)

roles/django/defaults/main.yml
```diff
@@ -33,4 +33,5 @@ django_csp_enforce_uri: http://example.com/
 
 # media root owner
 media_user: "www-data"
-media_group: "{{ media_user }}"
\ No newline at end of file
+media_group: "{{ media_user }}"
+media_root: "/var/www/media/"
```
rlskoeser, line 37: we have the same default path set in multiple places; the other places weren't getting loaded in this playbook. suggestions for a better solution?

group_vars/prosody/vars.yml
```diff
@@ -68,7 +68,6 @@ solr_configset: cdh_ppa
 #     - "Python unit tests (3.6, 8.6, postgresql)"
 
 # default media root
```
acozine, line 70: Do you need the comment line now that the setting is gone?
```suggestion
```

roles/build_project_repo/tasks/main.yml
```diff
@@ -40,7 +40,7 @@
       # register repo_info for group_vars
       register: repo_info
 
-    - name: Get python package version
+    - name: Get the version for the python package/app being deployed
```
acozine, line 43: Do you need to discover this on each machine? Could you define it in vars instead?

roles/postgresql/tasks/backup_db.yml
```diff
@@ -1,25 +1,27 @@
...
-    name: "{{ application_db_name }}"
-    login_host: "{{ postgres_host }}"
-    port: "{{ postgres_port }}"
-    login_user: "{{ application_dbuser_name }}"
-    login_password: "{{ application_dbuser_password }}"
-    encoding: "UTF-8"
-    owner: "{{ application_dbuser_name }}"
-    state: "dump"
-    target: "{{ db_backup_path }}"
+  tags: db_backup
```
acozine, line 4: This needs to go under the `block` definition, if it applies to all tasks within the block.

acozine (commented):
Here's what jumps out at me on my first pass. I'd be happy to pair on refining the role/playbook if you'd like - I haven't read the role or playbook in detail yet, though.

roles/django/defaults/main.yml
acozine: The role defaults file is a great place for default values that should apply across environments and servers. Just remember that any other value for that variable will override the role default.

roles/build_project_repo/tasks/main.yml
rlskoeser: This is the version of the app that we're deploying, so it is specific to the app+install (which roughly equates to being specific to the machine, at least for our current setup). This is one of the things I was asking about when I joined ansible open house last week, it's a variable we use for paths but we have to run some code to determine it.

roles/postgresql/tasks/backup_db.yml
rlskoeser: I tried that and it doesn't work - I get a syntax error. I think it's behaving as expected?

However, I'm glad I took another look - when I tested this I did figure out that my recommended syntax for calling the playbook didn't need to specify the skip tags, since we're already limiting to specific tags.

rlskoeser:
@acozine thank you for review and feedback, I've done some cleanup based on your suggestions and questions. Is there anything you think is important to address before merging?  

There are larger problems with cdh-ansible that I'd like to address at some point, and would love to get your input on priority & order once we do an overview/deep dive on current state, but don't want to hold this up.

roles/build_project_repo/tasks/main.yml
acozine: Ah, so the variable you're establishing is not the python version, it's a deployment timestamp or something similar that changes with every fresh code deploy? If you're using a symlink  (like `current`) you might be able to define the correct path by specifying the symlink and `follow: true` instead of using this variable.

acozine (approved):
Looks good to merge. This makes the automation better, so let's not wait for more refinements. Happy to look at the variables issues together - maybe next week sometime?

roles/build_project_repo/tasks/main.yml
rlskoeser: thank you for explaining that more clearly! yes, that's exactly the case

in fact, I remember now that I set a playbook-specific variable that uses the `current` symlink for this playbook, since we're not deploying a new version and the version that would be deployed isn't guaranteed to match the current version with the settings we want to update

rlskoeser:
> Looks good to merge. This makes the automation better, so let's not wait for more refinements. Happy to look at the variables issues together - maybe next week sometime?

Thank you! That would be helpful, I've done some preliminary work in a different branch and I think I made some progress - I think that's when I tripped over / figured out the `set_fact`; it would be valuable to look at together.

PR #162, opened 2023-11-03
Update simrisk playbook and vars to deploy to pulcloud

rlskoeser:
- add pulcloud simrisk to hosts inventory and to simrisk playbook
- only run sandbox task on sandbox hosts
- configure simulatingrisk to run at app-specific root path on sandbox only

playbook works, and I was able to deploy to both sandbox and simrisk cloud at the same time 🎉 

## questions
- hosts config assumes/requires ssh config is setup for cloud with bastion host; should this be documented anywhere in ansible repo or readme?
- ansible complains that simrisk host == simrisk group; suggestions?

### DIFF
--- a/group_vars/sandbox/simrisk.yml
+++ b/group_vars/sandbox/simrisk.yml
@@ -14,13 +14,15 @@ gitref: '{{ ref | default("main") }}'
 
 # use python 3.9
 python_version: "3.9"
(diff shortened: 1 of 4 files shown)


kayiwa (approved):
Oh! Cool @rlskoeser

Issue #164, opened 2023-11-15
template files should end with .j2


rlskoeser:
applies to local settings templates in django role; check for others

rlskoeser:
implemented in #166

PR #166, opened 2023-11-15
Use .j2 extension on django local settings template files

rlskoeser:
per #164 

- rename template files so they use `.j2` extension
- update variables referencing template files
- update templates extending base template file'

(unintentionally introduced some formatting changes, I'm trying out a new editor)


acozine (approved):
Looks good. I would not remove the newline at the end of every file, but that's one of those whitespace things that every editor (and every person) does differently.

rlskoeser:
> Looks good. I would not remove the newline at the end of every file, but that's one of those whitespace things that every editor (and every person) does differently.

@acozine thanks for reviewing. I usually only apply aggressive whitespace cleanup and code formatting to specific files but was trying out the Zed editor for this and don't know yet how to customize the auto-format to make it file-type specific. I'm not sure it's worth the time to undo the whitespace changes now, but maybe should establish pre-commit formatting hooks for this repo like I have for others so that it will be consistent even if different people and editors do things differently.

PR #167, opened 2023-12-11
Adjust steps that are failing on bionic VMs

rlskoeser:
These are the fixes I needed to run existing playbooks on bionic machines.

Any reason not to merge these into main so everyone is using them?


quadrismegistus (approved):
Looks great! thanks for incorporating these

PR #171, opened 2023-12-20
Cleanup apache2

rlskoeser:


rlskoeser:
tested running cdhweb qa playbooks (uses compressor) and prosody and shxco qa; running into errors with the other two (solr reload, webpack build), but they don't seem related to the changes in this pr

PR #172, opened 2023-12-22
geniza jammy upgrade

kayiwa:
- add new jammy hosts to inventory
- add new media_root location to nfs file share


rlskoeser:
superseded by #173

PR #173, opened 2024-01-04
jammy uggrade for shxco and geniza

rlskoeser:
- add new common fonts task to django role
- configure shxco to use fonts task; add vaulted font archive file
- configure cdhweb to use new fonts task and remove cdhweb_setup role


kayiwa (approved):
OK! Thanks @rlskoeser

kayiwa (approved):
OK! Thanks @rlskoeser

PR #174, opened 2024-01-15
geotaste to PUL and sandbox

quadrismegistus:
Wanted to start this as a PR with a quick review before I hit play—would like to confirm (a) it won't mess up simrisk's config on sandbox and (b) the same question on the PUL server and whether the playbook as it is now has the capacity to connect to its own PUL machine or whether other preparatory steps need first be taken. Thanks for taking a look! :)

### DIFF
--- a/roles/build_project_repo/tasks/main.yml
+++ b/roles/build_project_repo/tasks/main.yml
@@ -42,6 +42,7 @@
 
     - name: Determine version of python app being deployed
       ansible.builtin.import_tasks: python_app_version.yml
+      run_once: true
       when:
         - python_app is defined
 


rlskoeser (commented):
This all looks fine to me. I think you should go ahead and run it and see what happens. If something goes wrong with simrisk on the sandbox, it's not a big deal for it to be down for a bit, we'll just need to poke at it and figure it out. I'm not sure we ever had both apps running at the same time, so it may require some tweaking.

There's a convenience thing you should add to the main sandbox vars, which is an alias to cd to the current version of your geotaste app and activate the environment.  I see now that I didn't add any comments, but you should be able to follow the pattern based on what's there for simrisk. This is equivalent to the conan shortcut on the other boxes - I set it up to handle multiple applications with different aliases.  (And we should think about where to document this; maybe the readme for the sandbox role.)

PR #176, opened 2024-01-19
Configure steps that should only run once

rlskoeser:
- add `run_once` to steps that should only be run once even if deploying to multiple hosts
- revise cron role so cron jobs will only be configured on a specified server

### DIFF
--- a/group_vars/geniza/vars.yml
+++ b/group_vars/geniza/vars.yml
@@ -123,6 +123,9 @@ solr_configset: geniza
 deploy_contexts:
   - "Python unit tests"
 
+# run cron jobs only on the first machine in the host group
+cronjob_host: "{{ groups['geniza_prod'][0] }}"
+
 # configure scripts to run as cron jobs
 crontab:
   - name: "{{ app_name }} reindex"
--- a/group_vars/geniza/vars.yml
+++ b/group_vars/geniza/vars.yml
@@ -124,7 +124,7 @@ deploy_contexts:
   - "Python unit tests"
 
 # run cron jobs only on the first machine in the host group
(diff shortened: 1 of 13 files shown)


acozine (commented):
Overall this looks good.

The tasks in `solr_collection` may not need `run_once` - they are already running with `delegate_to` which should only run on the specific delegated server. I haven't dug into your inventory, but if `{{ solr_server }}` is a single box, then those tasks already only run on that box. That said, it won't do any harm.

roles/configure_crontab/tasks/main.yml
```diff
@@ -1,6 +1,7 @@
 - name: Configure cron jobs
+  when: inventory_hostname == cronjob_host
```
acozine, line 2: Hm. I'm not sure this will always reliably run on the same server - depending on which strategy you're using and where this task comes in the play, it might run on different servers on different runs, though I'm not sure either way. If this task must always run on the same specific server, then you may need to designate that server a different way.

group_vars/geniza/vars.yml
```diff
@@ -123,6 +123,9 @@ solr_configset: geniza
 deploy_contexts:
   - "Python unit tests"
 
+# run cron jobs only on the first machine in the host group
+cronjob_host: "{{ groups['geniza_prod'][0] }}"
```
acozine, line 127: Comment above should have been on this line . . . where `cronjob_host` is defined. This uses runtime order, which may change from playbook run to playbook run.

acozine: Hm, no, I think I was wrong. This is a pattern, so it does use inventory order. I think it's fine, as long as you don't add hosts to the group before the host you want to use as the `cronjob_host`.

PR #179, opened 2024-01-31
Remove workarounds for deploy steps that were failing on bionic machines

rlskoeser:


kayiwa (approved):
OK! Thanks @rlskoeser

PR #180, opened 2024-02-01
Update replicate playbook for multiple hosts in groups

rlskoeser:
update the replicate playbook to account for multiple hosts in host groups (which we have since the upgrade to jammy)

- use run once on all steps
- assume source host is the first host in the group

@quadrismegistus would you try this branch? I got an error on the step when it tried to remove the old database because it's being accessed by other users, would like to know if happens for you also. (I'm wondering if the other user is the second vm...)

### DIFF
--- a/playbooks/replicate.yml
+++ b/playbooks/replicate.yml
@@ -22,6 +22,7 @@
     - name: Generate a database dump
       # use existing postgres backup task with customized backup path
       # (role ensures backup dir exists)
+      run_once: true
       include_role:
         name: postgresql
         tasks_from: backup_db.yml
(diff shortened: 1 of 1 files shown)


quadrismegistus (approved):
Looks great! Very useful to have and for it to keep up pace as our VM system grows.

rlskoeser:
I don't remember now why I left this open - I feel like there was something else I wanted to do before merging, but there's no documentation on what that was, and I can't remember now! We've been successfully using it for a couple of projects, so I'm going to go ahead and merge because we do need the changes in main.

PR #184, opened 2024-02-08
test the nobody nogroup

kayiwa:
nfs4 needs nobody and nogroup
this tests if the conan group and www-data will work with it


rlskoeser (approved):
I'm fine with this since it seems to work and solves our permissions problem, but can you add a comment somewhere with a note explaining why went this route?   (Does it rise to the level of an ADR ? Nick set that up for this repo but we haven't used it)

PR #192, opened 2024-03-04
Add MapTiler API key to geniza vault and settings

blms:
Adds a newly required API key to the vault and django settings for Geniza. Related to https://github.com/Princeton-CDH/geniza/issues/1494

EDIT: Now also includes a fix for a bug in `geniza_settings` where `runtime_env` was written as `runtime`


rlskoeser (approved):
Thanks for catching my mistake on the runtime env, and for adding thew new token to the vault!

PR #197, opened 2024-04-19
Update prosody configuration to use solr 9 servers

rlskoeser:


rlskoeser:
solr 9 upgrade is in production; merging this config change into main

PR #200, opened 2024-05-03
reinstates lpass integration, moves inventory into a directory for tower

acozine:
If we move the inventory into a directory instead of using the default `hosts` file, Tower can import the inventory hosts and groups without needing access to the vault password.

I hope this means that we can keep the lpass integration while also using Tower to run playbooks. Let's test to confirm.


acozine:
With this change-set, in Tower we get this error:

```
Using /runner/project/ansible.cfg as config file
/runner/project/bin/lpass_default.sh: line 5: lpass: command not found
[WARNING]: Error in vault password file loading (default): Vault password
script /runner/project/bin/lpass_default.sh returned non-zero (127): None
/runner/project/bin/lpass_geniza.sh: line 5: lpass: command not found
[WARNING]: Error in vault password file loading (geniza): Vault password script
/runner/project/bin/lpass_geniza.sh returned non-zero (127): None
ERROR! Vault password script /runner/project/bin/lpass_geniza.sh returned non-zero (127): None
```
So we may still need to remove the scripts to get Tower working.

acozine:
After the second commit, which comments out the ansible.cfg setting for the lpass integration, but leaves the scripts themselves, the inventory sync works again. I'll leave it to @rlskoeser to document how to use the scripts without that configuration.

rlskoeser (approved):

PR #203, opened 2024-05-17
Update deploy and nginx config to serve Derrida archive with python wayback

rlskoeser:
- refine nginx configurations and playbook for derrida web archive deploy
- add option to do a shallow git clone (tested and works for both cases)
- refactor datadog settings to put common setup in all vars


kayiwa (approved):
OK! Thanks @rlskoeser

PR #207, opened 2024-05-30
Remove old derrida playbooks and variables

rlskoeser:


acozine (commented):
These changes look fine, though looking at the new state of the inventory prompts two questions: 
Do we only run the 'crawl' in staging?
Why do we have 2 boxes in derrida_archive_staging but only one in derrida_archive_production?

rlskoeser:
> These changes look fine, though looking at the new state of the inventory prompts two questions: Do we only run the 'crawl' in staging? Why do we have 2 boxes in derrida_archive_staging but only one in derrida_archive_production?

Ah, these are good questions! 

1. The 'crawl' was used to generate the web archive that is deployed for playback on the other derrida archive servers / playbooks.  It was more of a worker machine than a true staging; it doesn't need a production equivalent, and it can probably be archived now. I'm not sure I would do it the same way if/when we do this again, but a worker machine for running the crawls is nice!

2. @kayiwa recommended adding a second box in production and I think agreed but we didn't get it set up yet.

rlskoeser:
I'm going to merge this, since the questions are not directly related to the changes in this PR.

PR #208, opened 2024-05-31
Remove all Springdale/RedHat specific tasks, conditions, and variables

rlskoeser:
Removes all of the old Springdale specific deploy steps and conditional Ubuntu/Springdale steps.


kayiwa (approved):
OK! Thanks @rlskoeser

PR #209, opened 2024-06-04
Add princetoncdh github deploy token as default

rlskoeser:
- add @princetoncdh token with deploy permissions to `all/vault.yml`
- use that vaulted token as default value for `github_token`
- update readme with additional information about github token / deploy setup
- add architecture decision record documenting this change


kayiwa (approved):
OK! Thanks @rlskoeser

PR #210, opened 2024-06-06
Vault variables instead of entire files

### LINKED ISSUE #213: convert ansible encrypted variable files to encrypt variable values only, for compatibility with tower
- piloted this for cdhweb deploy and tested with tower
- in-progress script to convert and check vaulted variables
- update deploy check to handle new style of vaulted vars

rlskoeser:
This update revises how we handle sensitive variables for compatibility with Ansible Tower

- all `vault.yml` encrypted variable files under `inventory/group_vars` are updated so only the values are encrypted, not the entire file
- no longer using the geniza vault id; everything is now encrypted with the default id
- custom python script to help with working with encrypted variables (encrypt, decrypt, check)
- updated pre-commit hook: uses the custom python script to check that group var vault variable values are encrypted
- updated readme documentation to describe current status and use of vaulted variables
- added an architecture decision record to document the change

resolves #213 

question: is it worth thinking about contributing back to ansible / ansible-vault so it works better for this approach to vaulted variables?  (i.e., adapting some of the logic in my python script)


kayiwa (approved):
OK! thanks @rlskoeser

PR #216, opened 2024-06-13
Update CDH web deploy for new Springload 4.0 version of the site

rlskoeser:
- update python to version 3.11
- update node to version 18
- remove django compressor role and step (no longer used)
- use webpack role
- update font archive with new set of fonts, and deploy to required path
- run pip install in the application directory to pick up springload devkit zip package


kayiwa (approved):
OK! Thanks @rlskoeser

PR #217, opened 2024-06-13
Add NFS configuration for TigerData mount point for PPA

rlskoeser:
- using NFS settings provided by Robert Knight for TigerData
- tested and working on both staging and production

updates 2024-07-01:
- refactor nfs steps into a separate task included in main setup role when nfs is enabled
- move tigerdata steps from prosody setup to an optional task in setup role; disabled by default, enabled for prosody and prodigy


kayiwa (approved):
OK! Thanks @rlskoeser

Issue #211, opened 2024-06-10
write ansible playbook for prodigy deploy


jerielizabeth:

rlskoeser:
We now have a playbook and role in cdh-ansible for deploying Prodigy, which is tested and running on staging with a placeholder recipe. It was more complex than I anticipated/hoped - updated this to medium/3 points.

I'm going to close this and create draft tasks for the next steps I'm aware of.

PR #220, opened 2024-07-10
Deploy updates for CDH web pre-production site

rlskoeser:
- add cdhweb_preproduction hostgroup
- add variables for cdhweb pre-production (new db name, same db user & password as cdhweb prod)
- customize a few production-only steps to also run when runtime_env is preproduction

also includes logic to ensure deploy user uid is consistent (cherry-picked from #211 )


inventory/group_vars/all/vars.yml
```diff
@@ -115,6 +115,7 @@ timezone: America/New_York
 ## replication playbook settings
 timestamp: "{{ ansible_date_time.date }}"
 deploy_user: "conan"
+deploy_user_uid: 1002
```
acozine, line 118: This seems a bit fragile. If you are comfortable with it, I do not object, but if we could find a way to make things work without mandating a UID, the system might be more resilient.

acozine (commented):
Overall this looks fine. I don't entirely understand how the deploys work, but since they do, I'm in favor!

inventory/group_vars/all/vars.yml
rlskoeser: I agree that it's weird. I wouldn't normally go this route, but TigerData permissions _require_ uid/gid mapping for the permissions configuration. Usually `conan` gets this uid anyway, but it depends on whether or not the PUL `deploy` user is created on our VMs. Our recent slack conversation helped me recognize that at least I could enforce the uid to be consistent since it is currently required for TigerData.

rlskoeser: I thought I put a comment somewhere about this - I'll add a note with this variable so we'll at least know why it's there

PR #221, opened 2024-07-10
Get NFS idmapd working with ansible

rlskoeser:
- Set nfs domain to princeton.edu 
- revise nfs-idmapd restart handler (unmask, enable, and restart both nfs-server and nfs-imdapd)

I think the domain in the idmapd.conf was the key piece we had wrong. I kept poking at this on other serves and would get it working manually (often working quickly because something was broken), but was never sure which step was the key. This time I got it manually working on cdh-web3 and then by trial and error got ansible to do the same thing on cdh-web4.


kayiwa (approved):
OK! Thanks @rlskoeser

PR #224, opened 2024-08-13
Playbook for shxco dataset exports

rlskoeser:
- new playbook to generate S&co data exports and copy them to a configured folder in google drive
- new setup task to install rclone; currently only enabled for S&co application

I have now (finally 😅) successfully run this in staging.

questions:
- should the rclone setup task be a role instead?
- the data export playbook has optional zip file logic with rsync command; remove or keep and use tags to control?

ref https://github.com/Princeton-CDH/mep-django/issues/827


kayiwa (approved):
thanks @rlskoeser

PR #227, opened 2024-08-19
Temporary fix to serve images locally

laurejt:


rlskoeser (approved):
I think this will work for this app, since we don't have versioned installs  (on our other apps the `install_root` changes when we deploy a new version).

If we have to live with this longer we would want to add some logic to have ansible create the path and make sure it has the right permissions for nginx to serve out the content, but seems reasonable to skip that for now

PR #228, opened 2024-08-27
Preliminary documentation folder with diagram for prodigy deploy setup

rlskoeser:


docs/README.md
```diff
@@ -0,0 +1,18 @@
...
+
+## Application-specific documentation
+- [Prodigy](applications/prodigy.md) 
+
+
+## General
+
+CDH ansibilized applications are deployed to virtual servers managed by PUL.
+
+Typically servers are running Ubuntu Linux, and we have two VMs for each environment (staging and production). For historical reasons and for brevity, our server names use `-test` for staging environments (e.g., for PPA the staging servers ase cdh-test-prosody1 and cdh-test-prosody2).
```
laurejt, line 11:
```suggestion
Typically servers are running Ubuntu Linux, and we have two VMs for each environment (staging and production). For historical reasons and for brevity, our server names use `-test` for staging environments (e.g., for PPA the staging servers are `cdh-test-prosody1` and `cdh-test-prosody2`).
```

docs/README.md
```diff
@@ -0,0 +1,18 @@
...
+
+We access PUL-managed servers using ssh and the `pulsys` (PUL Systems) account.
+Access is based on ssh keys associated with team member GitHub accounts; this is managed
+in [princeton_ansible variables](https://github.com/pulibrary/princeton_ansible/blob/main/group_vars/all/vars.yml#L115). If you need to update ssh keys on a particular server, you can use the 
+"Update pulsys user keys" template in Ansible Tower and limit to the appropriate servers.
```
laurejt, line 16: I'm not sure what this means. Where would I find this template?

laurejt (commented):
Generally looks good. Most of my comments are stylistic / spelling fixes.

docs/README.md
rlskoeser: Let me explain here and then you help me figure out what we should add to our documentation. I'm also asking if it's ok to link directly to templates.

Here's a link to the template I mean in Ansible Tower: https://ansible-tower.princeton.edu/#/templates/job_template/17/details

When you "launch" the template/playbook, there is an input box for "limit" which is equivalent to the `--limit` command when running a playbook locally, i.e. only run on the host or host group I specify rather than all the hosts/host groups listed in the playbook

rlskoeser:
Thanks for catching all the spelling errors.  🤦‍♀️ 

And for suggesting proper case / titles everywhere. I made the changes locally and added you as a co-author on the commit rather than doing through the GitHub interface, and tested the mermaid diagram with a preview before I committed the changes. (Might want to adjust the image files / tigerdata part of the diagram but it's accurate for our current configuration.)

docs/README.md
laurejt: Ah, so this is within the Ansible Tower application.

I think this is my first time navigating to [ansible-tower.princton.edu](https://ansible-tower.princeton.edu), but I can't access any Templates (within the Resources drop down). This is also true when trying to navitgate to the link you shared.

Perhaps this is a permissions issue?

jerielizabeth: From the demo at standup, to add new keys to the CDH group, modify https://github.com/pulibrary/princeton_ansible/blob/main/group_vars/all/vars.yml. Add to documentation :D

laurejt (approved):

PR #229, opened 2024-08-30
Make cdhweb preproduction hosts and variables production

rlskoeser:
This update moves old production host and variables to *oldproduction* and what were previously pre-production hosts and variables to production, since the new site is now live and running in production.  This is temporary, and the old production hosts and variables will be cleaned up after the transition.

I'm leery of breaking anything, but the mismatch in environment names and status (pre-production/production) in ansible tower was confusing to the contractor team. 

I successfully used this new configuration to replicate the new production site data to staging.

I haven't run any deploys to production. Do you recommend running it against one VM to check?


kayiwa (approved):
OK! Thanks @rlskoeser

Issue #185, opened 2024-02-08
enable id_mapping on the client


kayiwa:
nfs needs id_mapping enabled. In order to do this the client file at 

`/sys/module/nfs/parameters/nfs4_disable_idmapping` should be set to N 

this may need a copy file instead of a line in file.

rlskoeser:
I think I've finally figured out the last piece of the nfs + idmap puzzle, and why it was apparently working in production but not staging. The config mentioned above apparently does not matter (it's only used when some other configuration is set that we're not using (?!)). What does matter is the domain in the idmapd.conf file - and because of the way we're setting it (using `lineinfile`), we ended up with multiple domains, which confused things (or at least confused me).

I figured out that you can check cached nfs-idmap resolver keys by running:
```sh
sudo nfsidmap -l
```
example output:
```
3 .id_resolver keys found:
  gid:users@lib-fs-staging.princeton.edu
  gid:conan@lib-fs-staging.princeton.edu
  uid:conan@lib-fs-staging.princeton.edu
```

It seems that these are only loaded after you've listed the contents of the files on the mount point. When I ran it before doing that I didn't get any values.

The domain in the idmap file needs to match that domain - i.e., it needs to match the nfs server name and not use the `.princeton.edu` domain as I'd thought was necessary at some point.

PR #235, opened 2024-10-01
Configure nfs idmap so it works for BOTH staging and production

rlskoeser:
My comments on #185 were partially wrong - the configuration I had before _was_ correct for production.  I'm not sure why staging and production NFS servers behave differently in terms of domain, but this configuration works for both staging and production.

On a production server:
```{bash}
$ sudo nfsidmap -l
4 .id_resolver keys found:
  uid:www-data@princeton.edu
  user:33
  gid:conan@princeton.edu
  uid:conan@princeton.edu
```
  
On a staging server:
```sh
$ sudo nfsidmap -l
3 .id_resolver keys found:
  uid:www-data@lib-fs-staging.princeton.edu
  gid:conan@lib-fs-staging.princeton.edu
  uid:conan@lib-fs-staging.princeton.edu
```


kayiwa (approved):
OK! Thank you @rlskoeser

PR #238, opened 2024-10-03
Changes in prepartion for round two.

laurejt:
These settings don't seem to fully work yet. For some reason images are not being displayed.


rlskoeser (approved):
All the changes you've made look good to me.

The ppa image dir variable change looks right. I checked the staging server and the nginx config still has the old image path. Then I checked the prodigy setup role in cdh-ansible and I see that the nginx setup task is tagged a setup task, which means it is skipped by default. I think that's still a reasonable default, we just have to remember to use the `-t setup,all` when we need to make changes to the nginx config. If we find we're doing that a lot then we can remove the setup/never tags from that task.

PR #240, opened 2024-10-07
Configure HTR to use PUL ponyexpress mail relay

rlskoeser:
Mail now can be sent with these configurations without errors, but my test messages have not yet arrived.


kayiwa (approved):
OK! Thanks @rlskoeser

PR #247, opened 2024-11-15
adds molecule, README to build dependencies

VickieKarasic:
added FQDN in the tasks

### DIFF
--- a/roles/build_dependencies/README.md
+++ b/roles/build_dependencies/README.md
@@ -1,0 +1,35 @@
+Role Name
+=========
+
+This role builds dependencies for CDH applications. 
+
+Requirements
+------------
+
+
+Role Variables
+--------------
+
+
+Dependencies
+------------
+
+
+Example Playbook
+----------------
+
+Including an example of how to use your role (for instance, with variables passed in as parameters) is always nice for users too:
+
+    - hosts: servers
+      roles:
+         - { role: username.rolename, x: 42 }
+
+License
+-------
+
+BSD
+
+Author Information
+------------------
+
+An optional section for the role authors to include contact information, or a website (HTML is not allowed).
--- a/roles/build_dependencies/meta/main.yml
+++ b/roles/build_dependencies/meta/main.yml
@@ -1,0 +1,17 @@
+---
+galaxy_info:
+  role_name: build_dependencies
+  company: Princeton University Library
+  description: build_dependencies
+  author: pulibrary
+
+  license: MIT
+
+  min_ansible_version: 2.2
+
+  platforms:
+    - name: Ubuntu
+      versions:
+        - jammy
+dependencies: []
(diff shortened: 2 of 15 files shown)


roles/build_dependencies/README.md
```diff
@@ -0,0 +1,35 @@
...
+Including an example of how to use your role (for instance, with variables passed in as parameters) is always nice for users too:
+
+    - hosts: servers
+      roles:
+         - { role: username.rolename, x: 42 }
+
+License
+-------
+
+BSD
```
rlskoeser, line 30: we've been using apache 2

roles/build_dependencies/README.md
```diff
@@ -0,0 +1,35 @@
...
+
+
+Example Playbook
+----------------
+
+Including an example of how to use your role (for instance, with variables passed in as parameters) is always nice for users too:
+
+    - hosts: servers
+      roles:
+         - { role: username.rolename, x: 42 }
```
rlskoeser, line 25: is this boiler plate that hasn't been filled in yet? even if no reqs/deps shouldn't we say that explicitly?

roles/build_dependencies/meta/main.yml
```diff
@@ -0,0 +1,17 @@
+---
+galaxy_info:
+  role_name: build_dependencies
+  company: Princeton University Library
+  description: build_dependencies
+  author: pulibrary
```
rlskoeser, line 6: Should these be CDH? We could look at what we used before in the roles where we have added metadata - we should probably standardize if possible, although of course I want to give appropriate credit for PUL contributing to these ansible roles.

roles/build_dependencies/meta/main.yml
```diff
@@ -0,0 +1,17 @@
...
+
+  license: MIT
```
rlskoeser, line 8: our default has been apache 2

rlskoeser (commented):
I'm not sure how best to review these, is it useful/important to pull down and run molecule locally?  

It looks like the build dependencies readme hasn't been filled out, and I see inconsistent licenses in various places (the repo-wide license and our default has been apache 2).  

What is the `new_role` role for?

FYI we have a github actions for molecule tests but I disabled it because they were all failing. I suggest we re-enable it and then you all can update it and add roles to the list as you get them updated.

roles/build_dependencies/README.md
kayiwa: This is boiler plate.

kayiwa:
Heya @rlskoeser we are running the tests locally. (there's 3 of us) There's no need - that I see - for you to do this. At some point you will need to, though. I have a ticket to do an introduction to molecule that is still "in the oven"... I will put it in my DevOops repo and give it to y'all

PR #250, opened 2024-11-25
create a new deploy role

kayiwa:
consolidate the create and close tasks into one role
remove the create_deployment and close_deployment role

### DIFF
--- a/roles/deploy/README.md
+++ b/roles/deploy/README.md
@@ -1,7 +1,7 @@
 Role Name
 =========
 
-Role that deploys applications
+Role that deploys CDH applications
 
 Requirements
 ------------
@@ -17,6 +17,7 @@ github_token: "your_token"
 Dependencies
 ------------
 
+No dependencies
 
 Example Playbook
 ----------------
@@ -30,7 +31,7 @@ Including an example of how to use your role (for instance, with variables passe
 License
 -------
 
-BSD
+Apache2
 
 Author Information
 ------------------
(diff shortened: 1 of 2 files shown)


roles/deploy/README.md
```diff
@@ -0,0 +1,39 @@
+Role Name
+=========
+
+Role that deploys CDH applications
```
rlskoeser, line 4:
```suggestion
Role that creates and updates GitHub deployments when deploying CDH applications to reflect status and environment of application deployments.
```

roles/deploy/meta/main.yml
```diff
@@ -0,0 +1,16 @@
+---
+galaxy_info:
+  role_name: deploy
+  company: Princeton University Library
```
rlskoeser, line 4: CDH ? or both CDH & PUL ?

rlskoeser (commented):
Apologies for the delay on reviewing this.

Nick had proposed previously in this issue https://github.com/Princeton-CDH/cdh-ansible/issues/45 that these two tasks go into a role named `github` since the logic is specific to the github api / github deployments. What do you think about that? I think calling the role "deploy" is misleading / ambiguous.

I think the create deployment and close deployment steps need to be separate tasks - right now they are both in the main tasks. Does this allow us to run the create deploy at the beginning of a playbook and then close it at the end of the playbook? 

This fail task is littered throughout the other roles (unfortunately) - I'm excited that we're closer to fixing it, but it touches a lot of parts of cdh-ansible. Is there a way to stage it? Can we include the task in the new location for now and start updating roles and playbooks to use the create and fail tasks

kayiwa:
superceded by https://github.com/Princeton-CDH/cdh-ansible/pull/288

PR #255, opened 2025-02-11
use the solrAPI

kayiwa:
instead of using rsync we use solrcloud's API to upload a configset we zip up the file and save it to a tmp directory and ship it user curl


rlskoeser:
Closing since we no longer need this - resolved by #256 instead

PR #263, opened 2025-04-29
Update S&Co config for Django split settings

blms:
**Associated Issue(s):** Princeton-CDH/mep-django#849

### Changes in this PR

- Update path to local settings in `shxco/vars.yml`
- Update settings template to not rely on `os` import


rlskoeser (approved):
Thank you!

Issue #265, opened 2025-05-07
update S&co deploy to use python 3.12


blms:

blms:
Closed by merge of #266 and #267.

PR #267, opened 2025-05-07
Handle distutils/setuptools for python 3.12 (#265)

blms:
**Associated Issue(s):** #265

### Changes in this PR

- Ensure distutils package is only installed for python < 3.12, as it was [removed from python](https://docs.python.org/3/library/distutils.html) in that version.
   - Prevent error on S&Co:
     `fatal: [cdh-test-shxco1.princeton.edu]: FAILED! => {"changed": false, "msg": "No package matching 'python3.12-distutils' is available"}`

### Notes

- Is there a way I can test this before running it in tower? Should I just do the resync and then run off this branch before merging? I imagine we want to make sure it doesn't break other scripts using this task…
- Hopefully I got the ansible yaml syntax right, it's still new to me (had to look up how to ensure that `python_version` is compared correctly using [this test](https://docs.ansible.com/ansible/latest/collections/ansible/builtin/version_test.html))
- I also removed `requires_python2to3` as I don't think that's true anymore for S&Co, and it would lock setuptools to an older version.

### DIFF
--- a/inventory/group_vars/shxco/vars.yml
+++ b/inventory/group_vars/shxco/vars.yml
@@ -78,8 +78,6 @@ app_dependencies:
   - at # for 100 years twitter bot scheduled tweets
   - libmysqlclient-dev # for mysql installation during the migration
 
-# 2to3 support needed for installing rdflib-jsonld
-requires_python2to3: true
 # datadog app/service name
 datadog_app_name: cdh_shxco
 
--- a/roles/python/tasks/install_python.yml
+++ b/roles/python/tasks/install_python.yml
@@ -21,6 +21,14 @@
       - python3-setuptools
       - python{{ python_version }}
       - python{{ python_version }}-venv
+    state: latest
+    update_cache: true
+
+- name: Install distutils for python < 3.12
+  become: true
+  ansible.builtin.apt:
+    name:
       - python{{ python_version }}-distutils
     state: latest
     update_cache: true
+  when: python_version is version('3.12', '<')


rlskoeser (approved):

rlskoeser:
>     * Is there a way I can test this before running it in tower? Should I just do the resync and then run off this branch before merging? I imagine we want to make sure it doesn't break other scripts using this task…

I think there's a way to run from a different cdh-ansible branch in Tower but I'm not sure I've used it. I don't totally get how that works with syncing the project and the inventory - but it would be good to get clarity. Let's ask and document somewhere what options are.

>     * Hopefully I got the ansible yaml syntax right, it's still new to me (had to look up how to ensure that `python_version` is compared correctly using [this test](https://docs.ansible.com/ansible/latest/collections/ansible/builtin/version_test.html))

I haven't seen that syntax before! I think I've only compared numeric values. There are ways to write tests for ansible with molecule  - @kayiwa and others were working on adding some on our repo to help us out and also for training. Maybe this version check could be a good small testable case...

>     * I also removed `requires_python2to3` as I don't think that's true anymore for S&Co, and it would lock setuptools to an older version.

Great - if it's not true yet, then let's make it true because we shouldn't still be relying on it at this point!

blms:
@rlskoeser It actually worked to just use this field!

Screenshot 2025-05-08 at 11 34 03 AM

It is correctly skipped here:

Screenshot 2025-05-08 at 11 33 30 AM

But not in geniza, which is still on 3.9:

Screenshot 2025-05-08 at 11 49 47 AM

---

However, I'm now once again getting this error in S&Co. I guess I'm still missing somewhere else this needs to be updated? Maybe it needs to be manually managed on the server? Let me know if I should ask Francis/PUL.

Screenshot 2025-05-08 at 11 35 37 AM

rlskoeser:
@blms thanks for this - and great to know about the cdh-ansible branch.

Can you test manually one on of the two servers and figure out what is going on / what is needed to resolve?

Actually, is something misconfigured somewhere?
The pyproject.toml in develop has this:
> requires-python = ">=3.10"

The python version in cdh-ansible group vars for shxco is python 3.12

The `.python-version` file also has 3.12.10

Maybe there's something wrong with one of the dependencies that is causing this problem?

One more question: is this a fresh install, or is it a hash that is already present on the server and might have an existing python virtualenv with packages that were previously instlaled?

PR #268, opened 2025-05-12
Update crontab role to support weekday; update prosody vars to specify weekday correctly

rlskoeser:
### Changes in this PR

- configure_crontab role now supports weekday configuration in host vars
- prosody variables updated to specify numeric weekday

### Notes

- Fixed to correct a bug that was found in production - cron job was running every week day for the first of the month, instead of on Saturdays only as intended.
- Playbook has already been run with the crontab tag to update the cron job with the corrected configuration. I checked the resulting config with [crontab.guru](https://crontab.guru/#0_2_1-7_*_6)


kayiwa (approved):
OK! Thanks @rlskoeser

Issue #270, opened 2025-05-28
Add DocRaptor config to PPA


blms:
requires `DOCRAPTOR_API_KEY` and `DOCRAPTOR_LIMIT_NOTE` Django settings.

PR #271, opened 2025-05-28
Add DocRaptor to prosody settings (#270)

blms:
**Associated Issue(s):** #270

### Changes in this PR

- Add DocRaptor settings to ppa template, add `docraptor_limit_note` to ppa group vars

### Notes

- We still need to add an API key, presumably to the vault. I'm guessing these should be separate for prod and staging, and should both be connected in some way to CDH accounts?

### DIFF
--- a/inventory/group_vars/prosody/vars.yml
+++ b/inventory/group_vars/prosody/vars.yml
@@ -88,3 +88,5 @@ datadog_app_name: cdh_prosody
 # configure scripts to run as cron jobs
 crontab: []
 # currently production only
+
+docraptor_limit_note: "Limited to 5 PDFs per month."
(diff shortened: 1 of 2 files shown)


rlskoeser:
@blms I had a thought: can DocRaptor API key and usage text be configured in Wagtail settings, so it doesn't require developer intervention to update? I haven't done this myself but we have something that I think is a custom config setting on a different project here: https://github.com/Princeton-CDH/cdh-web/blob/c6c842311ada54862868054a70ea7509e62c9aa9/cdhweb/pages/models.py#L592-L606

Sorry I didn't think of this before asking you to add it to ansible.

rlskoeser:
No longer needed since we went with wagtail settings instead https://github.com/Princeton-CDH/ppa-django/pull/722

PR #273, opened 2025-05-30
Remove datadog task from playbooks where it was added

rlskoeser:
**Associated Issue:**  —

### Changes in this PR

- Removed PUL datadog task from all deploy playbooks that referenced it

### Question

- Is any manual cleanup needed for existing deploys where this was run in the past?

### Evaluation

- [ ] Review playbooks and confirm change is appropriate


kayiwa (approved):
OK! Thanks @rlskoeser

PR #274, opened 2025-06-25
migrate to uv

kayiwa:
can we use uv to manage ansible

**Associated Issue(s):** #

### Changes in this PR

- `pip install -r requirements.txt` is now `uv pip install -r requirements.txt`
- use `uv`
- use `uv` in GHA pipeline

### Notes

Install `uv`


rlskoeser (changes requested):
@kayiwa this PR includes files for the external supervisor role that are installed by ansible-galaxy and shouldn't be checked into git; there's a line in the `.gitignore` file to ignore them but maybe the path has changed with newer versions of ansible?

Please  remove the .ansible/roles files and add the new path to .gitignore

There are also a lot of formatting changes to the readme file that is making it hard to see what text you actually changed. Were those intentional or something an editor did?  It looks like we lost a lot of formatting that I think we want.

kayiwa:
Oh weird. I have removed the galaxy based role additions. My text editor has a YAML syntax checker doo-hickey. (sorry about that)

PR #285, opened 2025-07-30
Fix replicate permissions for file access

tanhaow:
### Changes: 

Update the permission mode of the files to “0644” to access the media and database dump.

Before the fix, files were only readable by Conan (likely with permissions of 0600 or 0640). However, we SSH as Pulsys instead of directly as Conan, which prevented us from using `scp` to transfer Conan’s files directly.

After the fix, the files now have permissions of “0644,” allowing both Conan and Pulsys users to read them directly.

### Main files:
1. `postgresql/backup_db.yml` - handles DB dump permissions
2. `replicate.yml` - handles media permissions

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -116,3 +116,4 @@ venv.bak/
 
 # editor configs
 .vscode
+/.ansible
--- a/playbooks/replicate.yml
+++ b/playbooks/replicate.yml
@@ -43,7 +43,6 @@
         path: "{{ media_root }}"
         dest: "{{ media_backup_path }}"
         owner: "{{ deploy_user }}"
-        mode: "0644"
       become: true
       become_user: "{{ deploy_user }}"
       when: media_files.matched|int != 0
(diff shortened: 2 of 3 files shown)


rlskoeser (changes requested):
Your changes to file permissions are in the production part of the playbook where we're initially creating the backups, but the permission we want to fix is on the staging server.   

The part we need to modify is within the section where we switch to the staging hosts on line 60. The relevant tasks are line 92 (sync database dump to destination host) and line 115 (sync media to destination host).   It might make sense to do a single permission change once after both have completed.  

The other thing I wondered about is making sure permissions are correct on the directory we put the backup files on the staging server. That looks ok to me (line 85).

rlskoeser:
@tanhaow this could be a good opportunity to make sure you are set up to run cdh-ansible playbooks locally (with vault, lastpass-cli, etc). Once you make the changes, would you test the playbook on your branch (ppa would be a good one to test since we don't have any active development), and confirm that the permissions on the backup files are what you expect?

tanhaow:
Thanks @rlskoeser for catching that! You're absolutely right. I was fixing permissions on the production side... 😅 I've moved the permission fix to the staging section as you suggested, and I will get Vault and LastPass-cli configured to set up local testing. Thanks for the guidance!

tanhaow:
Testing Result:

- [x] **Ran cdh-ansible `replicate` playbook on ppa** to backup latest database and media dump from production to staging
- [x] **SSH'd to staging** and verified backup files are present: `ls -la /tmp/migrate/`
- [x] **Downloaded files from staging server to local machine** using `scp pulsys@cdh-test-prosody1.princeton.edu:/tmp/migrate/2025-08-05_cdh_ppa.sql ~/Downloads/`
- [x] **Verified no password prompt** - SSH keys working correctly and pulsys user can read files directly

--> ✅ No intermediate copy step required; direct download from pulsys user works as expected.

image

rlskoeser (approved):
Great! Thanks for revising. Thanks also for testing and reporting back, everything looks good.

Issue #46, opened 2020-11-25
automatically clean up old deploys


rlskoeser:
**Is your feature request related to a problem? Please describe.**
Currently, the deploy scripts leave all past deploys sitting in the directory.

**Describe the solution you'd like**
Old deploys should automatically be cleaned up when the deploy finishes. Maybe keep the last 3 based on date (most recent)? Should always preserve current + previous.

See for one possible solution https://www.future500.nl/articles/2014/07/thoughts-on-deploying-with-ansible/

**Describe alternatives you've considered**
Could be cleaned up manually or by a cron job, but seems better to make it part of the deploy.

PR #287, opened 2025-08-13
Automate clean up old deploys

tanhaow:
## Implements automatic cleanup of old deployment directories

Related issue:  #46 

### How it works:

The `cleanup_deploys` role is included into the `finalize_deploy` role and runs automatically as an async job after deployment finishes (fire-and-forget with 5-minute timeout).
The `cleanup_deploys` role does the following: 
  - finds all deployment directories using pattern matching
  - protects `current` and `previous` symlinks for rollback capability
  - keeps the 3 most recent deployments by default (configurable via `deploy_keep_count`), plus `current` and `previous`
  - removes old deployments 

### Changed files:

**New role:**
- `roles/cleanup_deploys/tasks/main.yml` - Core cleanup logic
- `roles/cleanup_deploys/defaults/main.yml` - Default configuration

**Test suite:**
- `test_cleanup.yml` - (Question: Is there a better place to put the tests?)


test_cleanup.yml
```diff
@@ -0,0 +1,105 @@
...
+      - "3.0.0-keep1"
+      - "3.1.0-keep2"
+      - "3.2.0-keep3"
+      - "4.0.0-previous"
+      - "4.1.0-current"
+
+  tasks:
+
+    - name: "Create test directory"
+      file:
```
kayiwa, line 33: these are not blockers. Can we use ansible fqrn so instead of `find` `ansible.builtin.find`

kayiwa (commented):
thanks @tanhaow this looks good. My comments are on the fully qualified role names.

rlskoeser (changes requested):
Thanks for working on this. Seems like you're making good progress on getting familiar with ansible!

How did you test this? I tried running a playbook in check mode and got an immediate failure. (I'm not sure if our playbooks work in check mode as we haven't used it, but would be nice in this case!). The test playbook is a clever solution and looks like it confirms the expected behavior works (I was able to run it), but doesn't match how the new role is being called.  Maybe we can get help from PUL turning the test playbook into a molecule test. 

I don't think this should be a new role; I think it should be a task (or possibly a handler) in another role - it seems reasonable to include within finalize_deploy for now; we have work in progress (but stalled) to consolidate some of the roles, but this shouldn't have to wait on that.

This needs changes but I think we need input from PUL ops team:
- decide whether it should be a task or a handler (this will determine where it gets moved and how it is called)
- how to adapt your test playbook into a molecule test

rlskoeser (changes requested):
Documenting the changes we agreed on in conversation with @kayiwa : 

- move the logic you have now in `cleanup_deploys/tasks/main.yml`  into a handler in the existing `finalize_deploy` role
   - use the ansible-galaxy init role command to add missing directories and files to the `finalize_deploy` role
   - update the `finalize_deploy` role main task to notify your new handler after the symlinks are updated
 - split out your `test_cleanup` playbook into a molecule test with a single scenario - the part that creates the fake deployments should be  in the `prepare` step and the steps you have to confirm expected results should be in the `verify`
    - re-enable the github actions molecule test workflow, add finalize_deploy, and comment out all other roles
 - test the cleanup handler manually at least once using an existing playbook deploy to  on a staging server

If the molecule part is hard enough that it and becomes a blocker for completing the work, we can split that out as a later task - but since you've already done such great work thinking about how to test the logic, I'd love to see it working and automated.

tanhaow:
Staging testing complete! - I tested the cleanup functionality on PPA and CDH Website staging servers. Cleanup worked on both servers, deleting deployments from many old versions down to exactly 5 directories while preserving the `current` and `previous` symlinks. 

```
=== CDH Web2 Deployment Count ===
5
=== Symlinks ===
lrwxrwxrwx  1 conan    users      28 Sep  9 21:43 current -> /srv/www/cdhweb/4.2.0-602138
lrwxrwxrwx  1 conan    users      28 Sep  9 21:43 previous -> /srv/www/cdhweb/4.2.0-23325c
=== Latest Deployments ===
lrwxrwxrwx  1 conan    users      28 Sep  9 21:43 previous -> /srv/www/cdhweb/4.2.0-23325c
lrwxrwxrwx  1 conan    users      28 Sep  9 21:43 current -> /srv/www/cdhweb/4.2.0-602138
drwxr-xr-x 13 conan    users    4096 Sep  9 21:42 4.2.0-602138
drwxr-xr-x 13 conan    users    4096 Aug 20 00:04 4.2.0-23325c
drwxr-xr-x 13 conan    users    4096 Jul 30 23:30 4.2.0-dcc2ea
```

PS: I noticed PPA's staging server 1 (prosody1) failed due to a missing `ecco_marc.vault` file.

rlskoeser (approved):
Great work. Let's merge! 🚀

tanhaow:
@kayiwa That's good to know. I wish I wish I knew it earlier. I had to log into Vault every time before I ran the test locally. :)
I also update it to use a separate SSH for testing.

PR #308, opened 2026-02-25
make cdh-ansible opinionated

### LINKED ISSUE #307: Update Ansible Configuration: Replace Removed community.general.yaml Callback Plugin Description:
When attempting to run an Ansible playbook with the following flags:

```sh
ansible-playbook -t setup,all \
  --skip-tags gh_deploy \
  --limit geniza_staging \
  --check \
  --diff \
  --vault-id default@bin/lpass_default.sh
```

The command fails with the following error:

```sh
[ERROR]: The 'community.general.yaml' callback plugin has been removed. The plugin has been superseded by the option `result_format=yaml` in callback plugin ansible.builtin.default from ansible-core 2.13 onwards. This feature was removed from collection 'community.general' version 12.0.0.
We are running Ansible 2.20.2, which is a recent version and not the cause of the issue. The root cause is a deprecated (now removed) callback plugin that must be replaced with the built-in equivalent.
```

### Expected Behavior

- Playbook runs in check/dry-run mode with YAML-formatted output, as intended.
- No errors related to callback plugins are raised.

### Actual Behavior

- Command fails immediately due to missing community.general.yaml callback.
- No output is generated.

### Root Cause
The `community.general.yaml` callback plugin was removed in **community.general collection v12.0.0.** It has been superseded by the built-in `ansible.builtin.default` callback with the `callback_result_format = yaml` setting.

kayiwa:
this change is necessary to allow us to use the same version of ansible on the ansible-automation platform. It addresses a problem where a developer has unable to successfully run a playbook because of a newer version of ansible was installed

**Associated Issue(s):** resolves #307 

### Changes in this PR

requirements.txt

### Notes

make sure your version of ansible is what is on pyproject and requirements.txt

```sh
…~/projects/pulibrary/cdh-ansible  i307_opinionated_ansibl… !2                                                                   impure 08:11:10 ❮ uv run ansible --version
ansible [core 2.19.4]
  config file = /Users/kayiwa/projects/pulibrary/cdh-ansible/ansible.cfg
  configured module search path = ['/Users/kayiwa/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /Users/kayiwa/projects/pulibrary/cdh-ansible/.venv/lib/python3.12/site-packages/ansible
  ansible collection location = /Users/kayiwa/.ansible/collections:/usr/share/ansible/collections
  executable location = /Users/kayiwa/projects/pulibrary/cdh-ansible/.venv/bin/ansible
  python version = 3.12.1 (main, Jan  7 2024, 23:31:12) [Clang 16.0.3 ] (/Users/kayiwa/projects/pulibrary/cdh-ansible/.venv/bin/python)
  jinja version = 3.1.6
  pyyaml version = 6.0.2 (with libyaml v0.2.5)
  ```


tanhaow (changes requested):
Thank you for making cdh-ansible opinionated and reproducible, @kayiwa!

I noticed there a inconsistency between `requirements.txt` and `pyproject.toml` that needs to be fixed.

We may consider remove `requirements.txt` and rely on `uv` + `pyproject.toml` since our Ansible roles already support `pyproject.toml`. This would eliminate version sync issues like this one and reduce maintenance burden. It can also simplify the setup process (it seems like the `pip`/`uv` dual options caused some confusion during @mabdellatif88's setup). -- But this is out of scope for this PR, but we can save it for future enhancement. What do you think, @rlskoeser?

PR #338, opened 2026-04-07
[run_webpack] validate and atomically publish stats

### LINKED ISSUE #336: `run_webpack` task fails with "webpack: not found" on staging/QA environment
The Ansible task `run_webpack : Run webpack in staging/QA mode` fails on host cdh-test-geniza2.princeton.edu with exit code 127, indicating that the webpack command is not found in the execution environment.

## Error Details
Failing Task: run_webpack : Run webpack in staging/QA mode

Error Output:

```text
fatal: [cdh-test-geniza2.princeton.edu]: FAILED! => {
  "changed": true,
  "cmd": "npm run build",
  "delta": "0:00:00.336642",
  "end": "2026-04-02 20:25:58.242311",
  "msg": "non-zero return code",
  "rc": 127,
  "start": "2026-04-02 20:25:57.905669",
  "stderr": "sh: 1: webpack: not found",
  "stdout": "\n> build\n> webpack --mode=production"
}
```
Secondary Error:

```text
TASK [run_webpack : include_tasks] fatal: [cdh-test-geniza2.princeton.edu]: FAILED! => {
  "reason": "Could not find or access '/runner/project/playbooks/roles/create_deployment/tasks/fail.yml' on the Ansible Controller."
}
```

## Steps to Reproduce
Run Ansible playbook against cdh-test-geniza2.princeton.edu host

Playbook executes run_webpack role

Task attempts npm run build which calls webpack

## Expected Behavior
Webpack should be available either:

Installed globally on the target host, or

Available via node_modules/.bin/webpack from a local installation

## Actual Behavior
The shell cannot locate the webpack command, causing the build to fail.

kayiwa:
Passenger/Django can read webpack-stats.json while webpack is still writing it, leading to intermittent WebpackLoaderBadStatsError and 502 responses. Our previous implementation also placed stats validation/publish steps in the rescue path where they would never run after a hard fail.

This change adds post-build validation of webpack-stats.json and rewrites it atomically in-place to avoid partial reads. The rescue path is kept for logging and a clear failure.

Side effects: adds a small amount of extra work after each webpack build (JSON parse + cp/mv), but improves deploy reliability.

**Associated Issue(s):** resolves #336 

### Changes in this PR

- Add post-webpack validation of `sitemedia/webpack-stats.json` and atomically rewrite it in-place to prevent Passenger/Django from reading a partially-written stats file.
- Keep the existing failure handling in `run_webpack`, but move stats validation/publish out of the rescue path so it runs on successful builds.

### Notes

- This addresses intermittent `webpack_loader.exceptions.WebpackLoaderBadStatsError` observed in Passenger/Nginx logs when requests hit the app while webpack is writing the stats file.
- `/var/www/geniza` is a symlink to the active release under `/srv/www/geniza/...`, so atomically updating the stats file under `npm_install_path` updates what Passenger reads without Geniza-specific path logic.

### Reviewer Checklist
_Include **discrete** checks that should be done by the reviewer beyond looking through
code and/or file changes. Note that this check list will correspond to tasks within
the PR overview page._

- [x] Deploy to Geniza test and confirm Passenger no longer emits `WebpackLoaderBadStatsError` during/after deploy (watch `/var/log/nginx/error.log`).
- [x]  Verify `{{ deploy }}/sitemedia/webpack-stats.json` is valid JSON and non-empty after `run_webpack` completes, and that requests succeed without 502s during the build window.

### DIFF
--- a/playbooks/geniza.yml
+++ b/playbooks/geniza.yml
@@ -31,8 +31,3 @@
     - configure_crontab
     - finalize_deploy
     - close_deployment
-  post_tasks:
-    - name: restart passenger
-      ansible.builtin.systemd:
-        name: nginx
-        state: restarted
--- a/playbooks/geniza.yml
+++ b/playbooks/geniza.yml
@@ -31,3 +31,8 @@
     - configure_crontab
     - finalize_deploy
     - close_deployment
+  post_tasks:
+    - name: restart passenger
+      ansible.builtin.systemd:
+        name: nginx
+        state: restarted
--- a/roles/run_webpack/tasks/main.yml
+++ b/roles/run_webpack/tasks/main.yml
@@ -21,7 +21,7 @@
           Missing webpack prerequisites in {{ npm_install_path }}.
           package.json={{ pkg.stat.exists | default(false) }},
           node_modules/.bin/webpack={{ webpack_shim.stat.exists | default(false) }}.
-          Ensure build_npm ran for this release and npm_install_path points at /srv/www/<app>/current.
+          Ensure build_npm ran for this same npm_install_path.
       when: not (pkg.stat.exists | default(false)) or not (webpack_shim.stat.exists | default(false))
 
     - name: run_webpack | Run webpack in staging/QA mode
@@ -43,6 +43,24 @@
         chdir: "{{ npm_install_path }}"
         executable: /bin/bash
       when: runtime_env in ["production", "preproduction"]
+
+    # Post-build: validate stats + atomically rewrite it in-place to avoid
+    # partial-read races by Passenger/Django.
+    - name: run_webpack | Validate webpack stats JSON
+      ansible.builtin.command: >
+        python3 -c 'import json; json.load(open("{{ npm_install_path }}/sitemedia/webpack-stats.json")); print("ok")'
+      changed_when: false
+
+    - name: run_webpack | Atomically publish stats file (same dir)
+      ansible.builtin.shell: |
+        set -euo pipefail
+        src="{{ npm_install_path }}/sitemedia/webpack-stats.json"
+        tmp="{{ npm_install_path }}/sitemedia/webpack-stats.json.tmp"
+        cp -f "$src" "$tmp"
+        mv -f "$tmp" "$src"
+      args:
+        executable: /bin/bash
+      changed_when: true
 
   rescue:
     - name: run_webpack | Show webpack failure output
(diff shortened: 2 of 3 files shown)


playbooks/geniza.yml
```diff
@@ -31,3 +31,8 @@
     - configure_crontab
     - finalize_deploy
     - close_deployment
+  post_tasks:
+    - name: restart passenger
+      ansible.builtin.systemd:
+        name: nginx
+        state: restarted
```
tanhaow: May I ask why we need to restart the nginx sever? 🤔

tanhaow (changes requested):
Hi Francis! Thank you for working on this. I reviewed this PR yesterday but forgot to submit the review... The two items in the checklist passed. But today when I take another look, I actually found there's a problem here.

I believe the racing problem is actually happening at the `webpack` layer, not the `Ansible` layer:

`webpack` is writing the stats file word by word, and `Passenger `might start reading it before `webpack` finishes writing. By the time `Ansible` runs its `cp` + `mv`, `webpack` has long since finished, so the atomic rename happens on an already-complete file and doesn't really protect against anything. I think the fix actually needs to happen inside `webpack` itself (in the `geniza` repo), so it writes to a temp file first and then renames it into place in one instant. 🤔

The checklist did pass though. But I think it happened to pass because the race condition only triggers if a request hits the app at the exact moment `webpack` is writing the file, which is a narrow window and I got lucky.

Also, @rlskoeser suggested we wire @mabdellatif88 in since it's related to geniza. So I also requested him as a reviewer as well.

playbooks/geniza.yml
kayiwa: Heya @tanhaow When I ran it multiple times the errors only went away after I restarted nginx. I don't fully get the webpack/django bits to fix it.

tanhaow: I think this might be caused by the [caching](https://github.com/Princeton-CDH/geniza/blob/main/geniza/settings/components/base.py#L204C9-L204C23) in geniza's Django (which cached the incomplete `webpack-stats.json`). But I'm not very sure. We can look at it together later

kayiwa:
heya @tanhaow Thanks for the careful re-review—agreed. The race is at the webpack write itself; Ansible’s cp+mv runs after webpack finishes, so it can’t protect the window where Passenger might read a partially-written file. I updated this PR to remove the atomic rewrite and keep only a post-build validation (exists/non-empty/valid JSON) so we fail fast on broken output.

tanhaow (approved):
The new resolution (post-build validation) looks good to me. Thank you for working on this, Francis @kayiwa !

mabdellatif88 (approved):

PR #346, opened 2026-05-01
Fix vault_vars script to handle empty chunks; improve output formatting

### LINKED ISSUE #345: vault_vars script fails to encode a new variable in an existing vault file
**Describe the bug**
vault_vars.py script errors when trying to encrypt a new vaulted variable

**To reproduce**
1. Add a new variable to an existing vault file in plain text
2. Run vault_vars script to encrypt the new variable (e.g., `./bin/vault_vars.py encrypt inventory/group_vars/.../vault.yml` )

**Expected behavior**
- The new variable should be encrypted in the vault file; existing variables should be preserved as is.
- Decrypting the vaulted variable should result in the original value.

rlskoeser:
**Associated Issue(s):** resolves #345 

### Changes in this PR

- Update `vault_vars` script to handle empty chunks and improve output formatting

### Reviewer Checklist

- [x] Review code changes
- [x] Test behavior using the steps documented on #345 to confirm the bug is fixed

### DIFF
--- a/bin/vault_vars.py
+++ b/bin/vault_vars.py
@@ -15,7 +15,6 @@ import shutil
 from pathlib import Path
 
 import yaml
-import ansible.cli
 from ansible.constants import DEFAULT_VAULT_ID_MATCH
 from ansible.parsing.vault import VaultSecret, VaultLib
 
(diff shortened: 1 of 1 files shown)


tanhaow (approved):
Tested the fix locally as instructed (adding a new variable to an existing vault file), and the encryption completes without error. I've also tested decrypting to confirm the new var round-trips to its original value. Also, block comments are retained in the output, and the format looks good. Thanks for working on this and LGTM!

PR #347, opened 2026-05-01
Escriptorium production deploy

### LINKED ISSUE #341: Update eScriptorium deploy to support new production environment

tanhaow:
**Associated Issue(s):** resolves #341 

### Changes in this PR
- Add `htr_production/vars.yml` with production variables (DB name `cdh_htr`, URL `htr.cdh.princeton.edu`, TigerData media path)
- Add `htr_production/vault.yml` with encrypted secrets (Django secret key, DB password)
- Update `playbooks/escriptorium.yml` to support production via -e `runtime_env=production` (defaults to staging)

### Reviewer Checklist
- [ ]  Verify `htr_production/vars.yml` variables look correct 
- [x] Confirm vault file is properly encrypted


rlskoeser (changes requested):
Production configuration looks like it has a lot of overlap with staging.  Please revise so that common configurations used in both environments is set in one place 

- define an htr group in `inventory/all_hosts` with htr production and staging as children (look for `[prosody:children]`or similar for an example).
- make an htr folder for shared settings; copy `htr_staging/vars.yml` to `htr/vars.yml` as starting point and then move anything that is distinct into staging and production variables (they should be much smaller).

The new vault file decrypts ok, but it needs to be changed so the variable names are visible and only the values are encrypted (use the `vault_vars` script or the command-line option in the readme to encrypt the values.). We switched to that for Ansible Tower; details are [documentated in an architectural decision record](https://github.com/Princeton-CDH/cdh-ansible/blob/main/architecture-decisions/0005-vault-variables-instead-of-files.md).

tanhaow:
Thanks for your review, @rlskoeser! I have made the following changes:
- added `[htr:children]` group to host inventory
- extracted and moved shared vars to `htr/wars.yml`
- only encrypted the values but not the keys of vars in `vault.yml`

inventory/group_vars/htr_production/vars.yml
```diff
@@ -0,0 +1,12 @@
+---
+###
+# Production-specific variables for eScriptorium
+###
+
+application_url: https://htr.cdh.princeton.edu
+passenger_server_name: htr.cdh.princeton.edu
+
+# production uses TigerData, not NFS
+nfs_enabled: false
```
rlskoeser: We will still need NFS to share media between the two machines. I think I included this in the VM request.

inventory/group_vars/htr_production/vars.yml
```diff
@@ -0,0 +1,12 @@
...
+tigerdata_enabled: true
+media_root: /mnt/tigerdata/cdh/htr/media/
```
rlskoeser: We can't use TigerData for the entire media folder; it isn't performant enough and storage is project-specific. TigerData integration will be partial and will come later.

rlskoeser (changes requested):
Looks good! This is very close.  The vault variables look good now.

Please configure both of them to use NFS and TigerData NFS. We're not ready to use the TigerData mount point yet, but we can confirm that we have access from these VMs.

And use the same media root for both of them, the one we were using before.

inventory/group_vars/htr_staging/vars.yml
```diff
@@ -1,157 +1,7 @@
...
-# - from list of third party tools in escriptorium full install instructions
-app_dependencies:
-  - "libpython{{ python_version }}-dev" # must match python version
-  - netcat-traditional
-  - jpegoptim
-  - pngcrush
-  - libvips
-
-application_db_name: "cdh_test_htr"
-application_dbuser_name: "cdh_test_htr"
```
rlskoeser, line 100: We need to preserve these in staging vars

inventory/group_vars/htr/vars.yml
```diff
@@ -0,0 +1,143 @@
...
+python_app_path: "{{ deploy }}"
+python_venv_prompt: "escriptorium"
+# default requirements file is directly under the app; escriptorium has it nested
+python_requirements_file: "{{ django_app_path }}/requirements.txt"
+# npm package.json file is in subdirectory "front"
+npm_install_path: "{{ python_app_path }}/front"
+# dev dependencies required to run the build
+npm_install_mode: "development"
+# escriptorium build command is npm run build
+webpack_build_qa: "production"
```
rlskoeser: This looks like a staging configuration.... You may need to define the non-qa webpack build command, or maybe the default will work. Let's leave this for now, I'm not sure what we need to change yet.

rlskoeser (commented):
I commented on a few small things, but generally things look good. I think you are at the point where you should start trying out your changes. 

Go ahead and run the playbook on this branch to deploy to the new production VMs (do not run against staging - let's leave staging untouched until after the migration). 

You will probably find more small things you need to fix once you try to deploy (I usually do). Once this is working in production, then I think we can merge.

tanhaow:
@rlskoeser, I removed `python_app` and `python_app_version` from the shared vars `htr/vars.yml`, which you originally set in `htr_staging/vars.yml`, because they're causing `build_project_repo` role fail.

I believe they were valid at the time when you were setup the staging server, but after Francis's [latest refactor](https://github.com/Princeton-CDH/cdh-ansible/pull/303) on `build_project_repo` role, it now has a version detection step that requires the `__version__` attribute which eScriptorium doesn't have.

But because our `build_project_repo/tasks/main.yml` only detects version "when: python_app is defined"：

```
- name: Determine version of python app being deployed
  ansible.builtin.import_tasks: python_app_version.yml
  when: python_app is defined
```

So i removed `python_app` form vars to skip that failing step.

rlskoeser:
> @rlskoeser, I removed `python_app` and `python_app_version` from the shared vars `htr/vars.yml`, which you originally set in `htr_staging/vars.yml`, because they're causing `build_project_repo` role fail.
> 
> I believe they were valid at the time when you were setup the staging server, but after Francis's [latest refactor](https://github.com/Princeton-CDH/cdh-ansible/pull/303) on `build_project_repo` role, it now has a version detection step that requires the `__version__` attribute which eScriptorium doesn't have.
...

Your solution is ok as a temporary workaround, but we need to resolve the problem. (I don't _think_ this is the only app affected.)   It will be important in the long run to distinguish which version if eScriptorium we're deploying.  You will have to figure out how to revise the `python_app_version` task to restore the previous behavior (allow override with a variable) or some other solution that works with eSciptorium.

tanhaow:
> You will have to figure out how to revise the `python_app_version` task to restore the previous behavior (allow override with a variable) or some other solution that works with eSciptorium.

Sounds good. I'll make a new issue about this

rlskoeser:
@tanhaow I think you should fix #348 as part of completing the initial production deploy ansible work, since it impacts the deploy path and has implications for future deploys.  If the fix is not trivial we can revisit.

inventory/group_vars/htr/vars.yml
```diff
@@ -0,0 +1,143 @@
...
+repo_url: https://gitlab.com/scripta/escriptorium.git
+gitref: dev-0.14.2
+
+# repo variable is still needed for checkout file structure
+repo: scripta/escriptorium
+# name of django application
+python_app: "escriptorium"
+# explicitly set a python app version since we require one and
+# ES doesn't seem to have any versions documented anywhere
+python_app_version: "main" # ?!
```
rlskoeser: Let's set this to match the gitref version we're installing, since we plan to update to the 1.0 release at some point

inventory/group_vars/htr/vars.yml
```diff
@@ -0,0 +1,143 @@
...
+django_app: "{{ python_app }}"
+# django app & python requirements are nested under the app folder
+django_app_path: "{{ deploy }}/app"
+# prefix for activate shortcut in conan profile
+python_venv_path_prefix: "app/"
+
+# symlink for web application
+symlink: "{{ app_name }}"
+# use python 3.11; kraken doesn't yet support 3.12
```
rlskoeser, line 34: Let's add to our list of tasks to check if/when we can upgrade to a newer python

playbooks/escriptorium.yml
```diff
@@ -15,7 +17,7 @@
     - configure_logging # logging directory must exist before running django commands
     - django
     - escriptorium_setup
-    - supervisor
+    - role: supervisor
       tags:
         - supervisor
```
rlskoeser, line 22: Why did this change?

rlskoeser (commented):
I've signed off on the vault vars. I had a few other comments/questions, and I see that you need to update this from cdh-ansible main branch now.  If the npm build is working the way you have things configured, then you can respond and resolve my question about that configuration.


Here's my list for this PR:
- Fix/restore the python app version logic and configure a meaningful version and confirm the deploy works.
- Respond to and resolve my comments.
- Decide whether you want to merge this branch now or wait until after production migration is complete (could depend on the switchover plan and whether you want to use ansible to update staging)

inventory/group_vars/htr_production/vars.yml
tanhaow: NFS is now enabled in the shared vars; both staging and production inherit it (`media_root: /mnt/nfs/cdh/htr/media/`). The production-specific vars no longer needs to set it explicitly.

inventory/group_vars/htr_production/vars.yml
tanhaow: Agreed. So now TigerData is just enabled via shared vars, but \`media_root\` still points to NFS (`/mnt/nfs/cdh/htr/media/`). TigerData integration will come later.

inventory/group_vars/htr_staging/vars.yml
tanhaow: I've recovered these. Thank you for catching it!

inventory/group_vars/htr/vars.yml
tanhaow: Now we have both `webpack_build_qa` and `webpack_build_prod` in `vars` and both are set to "production" - this is because eScriptorium's `front/package.json` only defines one build script called `production` and there's no separate `dev/qa` mode. 
(The variable name `webpack_build_qa` is a bit misleading here since now the value is set to `production`, but since it's part of the shared role interface we can't rename it -- just noting it here in the comment...)

inventory/group_vars/htr/vars.yml
tanhaow: Good suggestion! `python_app_version` is now set to match the gitref. This help replace the previous workaround on the deploy path. Now the deploy path use the standard format like other CDh projects.

playbooks/escriptorium.yml
tanhaow: The previous syntax (`- supervisor` with nested `tags:`) was invalid YAML: a plain list item (- value) cannot have nested mapping keys. -- Can be verified with a YAML parser. The change here (`- role: supervisor`) converts the list item into a mapping so tags can be attached.

tanhaow:
@rlskoeser I re-ran the Ansible playbook for testing, and it successfully executed without issues (Though I still need to add a workaround for the mirror problem). This PR is now ready for your review. I suggest we merge this branch now to close the issue up instead of waiting until after the entire switchover process.

inventory/group_vars/htr/vars.yml
```diff
@@ -0,0 +1,160 @@
...
+# Git repository (on gitlab instead of github, so need to specify the full repo url)
+repo_url: https://gitlab.com/scripta/escriptorium.git
+gitref: dev-0.14.2
+
+# repo variable is still needed for checkout file structure
+repo: scripta/escriptorium
+# name of django application (used by build_project_repo)
+python_app: "escriptorium"
+# set version to match gitref since eScriptorium has no __version__ attribute;
+# build_project_repo will skip auto-detection when python_app_version is pre-set
```
rlskoeser, line 24: nice comment, thank you for documenting

inventory/group_vars/htr/vars.yml
```diff
@@ -0,0 +1,160 @@
...
+  - jpegoptim
+  - pngcrush
+  - libvips
+
+# Override the default postgres_apt_repository to include signed-by.
+# Ubuntu 22.04 (jammy) requires the signed-by option to avoid an apt conflict
+# when the keyring file already exists at the default path. Without this override,
+# the postgresql role writes a line without signed-by, which conflicts with the
+# existing signed-by entry and causes apt to fail with "Conflicting values set
+# for option Signed-By".
```
rlskoeser, line 103: Is there an issue we're tracking this on? Worth adding a link so we can update when it's resolved?

rlskoeser (approved):
Revisions look good to me! And I agree with you, it seems like it will be good to get these changes merged in.

inventory/group_vars/htr/vars.yml
tanhaow: Good suggestion! I've opened the issue #356 and mentioned it as a TODO in code comments.

PR #350, opened 2026-05-08
I349 tigerdata

### LINKED ISSUE #349: Fix Prodigy setup file copy from TigerData
The prodigy_setup role copies Prodigy task data and instruction files from TigerData using shell commands:

```yaml
cp "{{ prodigy_datafile_src }}" "{{ prodigy_datafile }}"
chmod 755 "{{ prodigy_datafile }}"
```

This is failing during Ansible runs even though the same files are accessible manually from both VMs.

### Problem

The copy tasks currently run with:

```yaml
become: true
```

That means the `cp` command executes as `root`, not as the `conan` user(?). TigerData/NFS-style mounted storage commonly restricts or maps root access differently from normal user access. As a result, a file may be readable by `conan` but not readable by root through Ansible.

This makes the automation unreliable even though the source path is valid:

```sh
/mnt/tigerdata/cdh/muse/phase-1/prodigy/notion-concept/debug-task1.jsonl
```

kayiwa:
**Associated Issue(s):** resolves #349

### Changes in this PR

- Fix Prodigy data and instructions file copies from TigerData by avoiding raw `cp`/`chmod` shell tasks.
- Copy remote TigerData files using Ansible modules with `remote_src: true`.
- Run TigerData readability checks and copy operations as the deploy user so Ansible matches the user context that can access the mounted files manually.
- Remove `setup`/`never` tags from core role tasks so normal role execution and Molecule exercise the same setup path.
- Add Molecule coverage for the `prodigy_setup` role using mock Prodigy source files instead of requiring TigerData in the test container.
- Add Molecule verification for install directories, copied files, `requirements.txt`, `prodigy.json`, generated nginx config, and `nginx -t`.
- Add a README documenting role variables, TigerData copy behavior, nginx setup, example usage, and Molecule commands.

### Notes

- The likely root cause was that the Ansible copy step ran with `become: true` as root, while manual access was tested as a normal user. TigerData may allow the deploy user to read the files while denying or mapping root differently.
- The Molecule scenario uses local fixture files to keep tests repeatable without requiring a TigerData mount.
- The role now has clearer failure messages when a Prodigy data or instructions source file is missing or unreadable by the deploy user.

### Reviewer Checklist
_Include **discrete** checks that should be done by the reviewer beyond looking through
code and/or file changes. Note that this check list will correspond to tasks within
the PR overview page._

- [x] Confirm the TigerData source file checks run as `{{ deploy_user }}`.
- [x] Confirm the copied Prodigy data and instructions files end up owned by `{{ deploy_user }}`.
- [ ] Run `molecule test` from `roles/prodigy_setup`.
- [x] Confirm `prodigy.json` contains the expected PostgreSQL configuration.
- [x] Confirm the generated nginx config validates with `nginx -t`.

### DIFF
--- a/.github/workflows/molecule_tests.yml
+++ b/.github/workflows/molecule_tests.yml
@@ -7,7 +7,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        role: [build_dependencies, build_project_repo, build_npm, deploy_user, finalize_deploy, opencv_from_source, postgresql]
+        role: [build_dependencies, build_project_repo, build_npm, deploy_user, finalize_deploy, opencv_from_source, postgresql, prodigy_setup]
     steps:
       - uses: actions/checkout@v4
 
--- a/inventory/group_vars/prodigy/vars.yml
+++ b/inventory/group_vars/prodigy/vars.yml
@@ -28,7 +28,7 @@ tigerdata_enabled: true
 # Copying files from TigerData
 prodigy_data_dir_src: "{{ tigerdata_mount_dir }}/muse/phase-1/prodigy/notion-concept"
 prodigy_datafile_src: "{{ prodigy_data_dir_src }}/notion-concept-tasks.jsonl"
-prodigy_instruct_src: "{{ tigerdata_mount_dir }}/instructions.html"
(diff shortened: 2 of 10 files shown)


laurejt (approved):
Looks good to me.

PR #357, opened 2026-05-18
Switch eScriptorium to Adroit and fix Celery worker concurrency

tanhaow:
**Associated Issue(s):** [#81](https://github.com/Princeton-CDH/htr2hpc/issues/81) [#82](https://github.com/Princeton-CDH/htr2hpc/issues/82)

### Changes in this PR

- Switch htr2hpc install to `adroit` branch so training jobs submit to Adroit instead of Della (Della folder structure changed and is no longer compatible)
- Split single Celery worker into four per-queue workers (`celery-main`, `celery-gpu`, `celery-live`, `celery-low-priority`) to fix concurrent recognition jobs hanging or queuing indefinitely
- Update `HPC_HOSTNAME` to `adroit.princeton.edu` in the settings template
- Add `CELERY_TASK_ROUTES` to route tasks to the correct Celery queues

### Notes

- These changes are based on fixes Christine made during beta testing (cmroughan/cdh-ansible commits c4d0edf, 4e14659, 69ff4a8) that were never merged into the main repo
- The Celery worker split is the key fix for the production issue where multiple simultaneous segmentation/transcription jobs hang or queue indefinitely
- Adroit also requires a Duo exemption — that needs to be requested separately

### DIFF
--- a/inventory/group_vars/htr/vars.yml
+++ b/inventory/group_vars/htr/vars.yml
@@ -140,11 +140,11 @@ supervisor_programs:
       killasgroup=true
       stopasgroup=true
 
-  # 2) celery worker
+  # 2) celery workers (split by queue to prevent concurrent jobs from hanging)
   # based on the command in the full local deploy
-  # celery -A escriptorium worker -l INFO
-  - name: "celery"
-    command: "{{ django_app_path }}/env/bin/celery -A escriptorium worker -l INFO"
+  # (docker compose starts multiple celery workers with different queues/priorities)
+  - name: "celery-main"
+    command: "{{ django_app_path }}/env/bin/celery -A escriptorium worker -Q default --concurrency 1 -l INFO"
     state: present
     configuration: |
       directory={{ django_app_path }}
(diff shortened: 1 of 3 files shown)


inventory/group_vars/htr/vars.yml
```diff
@@ -37,7 +37,7 @@ python_version: "3.11"
 node_version: "18"
 # install local customizations as an extra package
 python_extra_packages:
-  - git+https://github.com/Princeton-CDH/htr2hpc.git@develop#egg=htr2hpc
+  - git+https://github.com/Princeton-CDH/htr2hpc.git@adroit#egg=htr2hpc
```
rlskoeser: This needs to change. We should ideally be deploying from main or a tagged release. Could we use the existing gitref logic and you can override on the command line until we have a new release of htr2hpc?  (Should be documented in the readme but if it's unclear let's improve the docs.)

We should also think about whether we publish the package to pypi in future.

roles/django/templates/escriptorium_settings.py.j2
```diff
@@ -159,7 +175,7 @@ SOLR_CONNECTIONS = {
 # Extra app-specific configuration
 {% block extra_config %}
 # custom config for htr2hpc
-HPC_HOSTNAME = "della.princeton.edu"
+HPC_HOSTNAME = "adroit.princeton.edu"
```
rlskoeser, line 178: This seems ok for now, but let's think about shifting this to an ansible variable in future. (Although how we approach it may change in future, so we may want to hold off until we revise that.)

inventory/group_vars/htr/vars.yml
```diff
@@ -154,8 +154,59 @@ supervisor_programs:
...
+    configuration: |
+      directory={{ django_app_path }}
+      environment=DJANGO_SETTINGS_MODULE="escriptorium.local_settings"
+      autostart=true
+      autorestart=true
+      startretries=1
+      startsecs=1
+      redirect_stderr=true
+      stderr_logfile=/var/log/supervisor/celery-low-priority-err.log
+      stdout_logfile=/var/log/supervisor/celery-low-priority-out.log
```
rlskoeser, line 213: I don't love the duplication of these celery queues - another case where this is acceptable for now but we should look at setting up based on a list of queues and consolidate the common logic in future.

roles/escriptorium_setup/handlers/main.yml
```diff
@@ -2,7 +2,12 @@
 # handlers file for escriptorium
 - name: Restart celery
   become: true
-  ansible.builtin.command: "sudo supervisorctl restart celery"
+  ansible.builtin.command: "sudo supervisorctl restart {{ item }}"
+  loop:
+    - celery-main
+    - celery-gpu
+    - celery-live
+    - celery-low-priority
```
rlskoeser, line 10: This should ideally be based on a list from the host variables — it's hard to tell if they are all included the way things are set up now. That can be included in the later refactor.

rlskoeser (changes requested):
I've flagged some things for future improvement. The things that I would like changed before merging:
- fix the htr2hpc install to use `main` by default; in future you can override the default for staging to use develop
- add a co-author commit for Christine since she set up the celery queue configuration (AFAIK you mentioned her commits but you didn't cherry-pick them directly)

tanhaow:
@rlskoeser Thanks for your review! I've made the following changes:
- Replaced the hardcoded adroit branch with a gitref variable
- Added co-author credit to Christine

The two items flagged for future refactoring are tracked in [htr2hpc#88](https://github.com/Princeton-CDH/htr2hpc/issues/88) and [htr2hpc#89](https://github.com/Princeton-CDH/htr2hpc/issues/89).

inventory/group_vars/htr/vars.yml
```diff
@@ -36,8 +36,12 @@ python_version: "3.11"
 # nodejs version
 node_version: "18"
 # install local customizations as an extra package
+# Note: the standard `-e ref=GITREF` override controls the eScriptorium gitref only,
+# not the htr2hpc package. To deploy a specific htr2hpc branch or tag, use:
+#   -e htr2hpc_gitref=adroit
+htr2hpc_gitref: main
```
rlskoeser: Ah! I forgot we were using the existing `gitref` variable for the eScriptorium version. This is a good solution. Please update htr staging variables to set the the default `htr2hpc_gitref` to `develop` for staging.

Thanks for the comment here - we'll want to improve the ansible docs for this playbook as we update things to make it more production-ready (I also see the node version also looks quite old), but that can be done later.

rlskoeser (approved):
I like the gitref solution  — I forgot we were using then default one for the eScriptorium version. We may want to revisit in future, but this looks good.

Thanks for creating the new issues to document future improvements, let's make sure those get added to appropriate epics for tracking.

The only thing I think should be changed before merging is setting the default `htr2hpc_gitref` for staging to `develop` - that isn't strictly necessary now since we don't yet have a staging site, but seems best to while we have the context.

rlskoeser:
@tanhaow I forgot to mention: the tests you put in the reviewer checklist for this PR are really more acceptance-testing style checklist, I suggest you move them to the relevant issues so Christine can use them as she tests.

inventory/group_vars/htr/vars.yml
```diff
@@ -36,8 +36,12 @@ python_version: "3.11"
 # nodejs version
 node_version: "18"
 # install local customizations as an extra package
+# Note: the standard `-e ref=GITREF` override controls the eScriptorium gitref only,
+# not the htr2hpc package. To deploy a specific htr2hpc branch or tag, use:
+#   -e htr2hpc_gitref=adroit
+htr2hpc_gitref: develop
```
rlskoeser, line 42: @tanhaow I don't see an override anywhere (maybe I'm missing it from how I'm looking at the changes) - default for production should be main and default for staging should be develop.  It should follow the same pattern we use for the other deploys.   If you don't want to deal with it yet then I think the most important thing is that production defaults to main.

tanhaow: Ugh sorry about that! clearly not fully awake this morning. I've opened a follow-up PR [#363](https://github.com/Princeton-CDH/cdh-ansible/pull/363) to fix this. `htr2hpc_gitref` now defaults to `main` in the shared vars and is overridden to `develop` in the staging vars.

PR #361, opened 2026-05-23
cdh observability

kayiwa:
**Associated Issue(s):** related to https://github.com/pulibrary/princeton_ansible/issues/6234

### Changes in this PR

Added variables for otel_collector and Beyla roles
Imported otel_collector and Beyla roles from prancible

### Notes

log into our new [observability platform](https://signoz-staging.lib.princeton.edu/services/cdh-web?startTime=1779572896802000000&endTime=1779574696802000000) for log, metrics and traces


rlskoeser (changes requested):
Hi, @kayiwa, thanks for adding this, I'm sure it will be useful.

I tried logging in to the staging observability platform with the link you shared, but when I try with the SSO link I get a Google auth error screen: "Access blocked: Authorization Error."  Can you let me know what I'm missing or doing wrong?

Regarding the new configuration - it looks like there is a lot of duplication. Could we put the common settings all in `inventory/group_vars/staging/vars.yml` and with enabled default to false and then each application that uses it can set enabled to true and define the service name?  (And update the operator value options to use the service name variable)

rlskoeser (approved):
Thanks for revising to avoid duplication, looks much better now.

I figured out my problem with logging in - something got blocked by my noscript plugin (I didn't see a way to ignore/bypass). I was able to log in with a different browser and can see charts for activity.  Looks useful!

PR #364, opened 2026-05-29
Fix nginx restart timing in supervisor role

tanhaow:
**Associated Issue(s):** resolves princeton-cdh/htr2hpc#93

### Changes in this PR
- Replace `meta: flush_handlers` in the supervisor role with an explicit unconditional `supervisord` restart task
- Remove all `notify: restart supervisor` from supervisor role tasks since the restart is now unconditional

### Notes
- Background: The supervisor role used `meta: flush_handlers` to force-restart supervisord immediately after config changes. This flushed *all* pending handlers including nginx, but at that point `finalize_deploy` hadn't yet updated the `/var/www/` symlink. After `finalize_deploy` updated the symlink and re-notified the nginx handler, Ansible wouldn't re-fire it — so Passenger kept serving the old code. Replacing `flush_handlers` with a direct supervisord restart task fixes the timing for all playbooks that use the supervisor role, not just `escriptorium.yml`.

### Checklist for myself
- [x] Run playbook against **production** 
- [x] Confirm `nginx` restart task handler got trigger at the end (not early)
- [x] Confirm `nginx` symlink points to latest deploy

### Reviewer Checklist
- [x] Review the code change

### Follow-up acceptance testing
Confirm training jobs work without manual nginx restart after deploy. [#98](https://github.com/Princeton-CDH/htr2hpc/issues/98)

### DIFF
--- a/playbooks/escriptorium.yml
+++ b/playbooks/escriptorium.yml
@@ -31,10 +31,11 @@
     - name: Restart supervisor apps to load any python/django configuration changes
       ansible.builtin.command: "sudo supervisorctl restart {{ item.name }}"
       loop: "{{ supervisor_programs }}"
-    - name: Notify handler to restart nginx
-      ansible.builtin.meta: noop
-      notify:
-        - Restart nginx
+    - name: Restart nginx to load new deploy
+      become: true
+      ansible.builtin.service:
+        name: nginx
+        state: restarted
   environment:
     # escriptorium local settings REPLACES default settings, must be specified
     # via env var for django manage commands to pick up
--- a/playbooks/escriptorium.yml
+++ b/playbooks/escriptorium.yml
@@ -31,11 +31,10 @@
     - name: Restart supervisor apps to load any python/django configuration changes
       ansible.builtin.command: "sudo supervisorctl restart {{ item.name }}"
       loop: "{{ supervisor_programs }}"
-    - name: Restart nginx to load new deploy
-      become: true
-      ansible.builtin.service:
-        name: nginx
-        state: restarted
+    - name: Notify handler to restart nginx
+      ansible.builtin.meta: noop
+      notify:
+        - Restart nginx
   environment:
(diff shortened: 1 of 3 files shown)


rlskoeser:
@tanhaow can we get rid of the supervisor meta call that is flushing the handlers too soon?  This feels like we are working around the problem rather than solving it properly, would be great if we could fix it in the role for all playbooks.

It might also be good to get a quick review from Francis or someone from the PUL devops team.

tanhaow:
@rlskoeser Great suggestion! If I understand it correctly, the `flush_handlers` was added there in the first place to restart `supervisord` after config changes, but the side effect was that it also flushed the `nginx` handler too early. So here I've replaced it with a service task that restarts `supervisord` directly, which no longer needs the `flush_handler`. I've also updated the PR description, reverted the `escriptorium.yml` workaround, and requested a review from @kayiwa.

rlskoeser (commented):
Will be good to get @kayiwa's input - I'm not sure about these changes, I think it would be better to keep the handler and just drop the flush changes. 

@tanhaow have you run the playbook with these changes to test behavior?

rlskoeser:
@tanhaow regarding your review checklist: I don't think anyone but you should be running this playbook to test the changes. Have you done that already to confirm that these changes do what you expect?

The last task looks like acceptance testing for Christine and should be on an issue rather than a PR. (I do think it's valuable for her to confirm independently from you that the changes fix the problem she identified.)

tanhaow:
@rlskoeser I updated the code per your suggestion (keeping the handler and removing the flush_handlers call) and verified the fix by running the playbook. I can confirm that the nginx handler now fires after the symlink is updated rather than early. I've also added acceptance testing to the issue and will request Christine's test after the PR got merged and the release is ready.

rlskoeser (approved):
Thanks for revising and confirming - I'm glad that this simple change is all we need.

(It probably never worked correctly before, since we were restarting before the new changes were live.)

PR #366, opened 2026-06-03
Fix ineffective nginx restart for htr2hpc

tanhaow:
**Associated Issue(s):** resolves #princeton-cdh/htr2hpc#102

### Changes in this PR
- Replace ineffective `meta: noop` + `notify` with an explicit service restart task at the end of `escriptorium.yml`
- The `meta` tasks do not support `notify`, so the `nginx` handler was silently never triggered after escriptorium patches were applied.

### Notes
- This bug was already exist and non-functional before the PR #364 fix.
- Verified the current fix is working when deployed Chrsitine's `feature/kraken6` and confirmed nginx restarted at the end of the playbook without requiring a manual restart.

### Reviewer Checklist
- [x] Review the code

### DIFF
--- a/playbooks/escriptorium.yml
+++ b/playbooks/escriptorium.yml
@@ -33,10 +33,10 @@
       loop: "{{ supervisor_programs }}"
       # skip absent programs because they have already been removed and cannot be restarted
       when: item.state | default('present') != 'absent'
-    - name: Notify handler to restart nginx
-      ansible.builtin.meta: noop
-      notify:
-        - Restart nginx
+    - name: Restart nginx after escriptorium patches are applied
+      ansible.builtin.service:
+        name: nginx
+        state: restarted
   environment:
     # escriptorium local settings REPLACES default settings, must be specified
     # via env var for django manage commands to pick up


rlskoeser (approved):
Thank you for finding this and finding a solution! I suspect it may be impacting other playbooks as well.

Is there no way to notify the handler? Or call it directly, maybe? I would prefer not to duplicate (not that it's much in this case).

tanhaow:
> Thank you for finding this and finding a solution! I suspect it may be impacting other playbooks as well.
> 
> Is there no way to notify the handler? Or call it directly, maybe? I would prefer not to duplicate (not that it's much in this case).

`meta:noop` with `notify` was only used in the eS playbook so this should not be impacting other playbooks.

Regarding the duplication, I think we could use `notify: restart web server` (the handler listens oin that string). But since nginx needs to restart unconditionally here rather than only when something changes, so a `service` task actually looks better to me than a handler. But I get your thoughts on avoiding duplication 🤔 - Happy to switch if you prefer!

rlskoeser:
@tanhaow you've already tested and confirmed this approach works, let's move forward with this.

PR #369, opened 2026-06-05
Revert "install otel and beyla on staging only"

kayiwa:
This reverts commit f474005db33e3bcd66f3a68a51131bcf73cfe0e5.

**Associated Issue(s):** resolves #

_If this PR has a deadline, note it here. Remove if there is no deadline._

**Deadline for review**: 

### Changes in this PR
_Include all key changes in this pull request_

- Change 1
- Change 2

### Notes
_Include any additional notes that will help in the reviewing of this pull request_

- Note 1
- Note 2

### Reviewer Checklist
_Include **discrete** checks that should be done by the reviewer beyond looking through
code and/or file changes. Note that this check list will correspond to tasks within
the PR overview page. If these changes has any impact on Geniza configuration or playbooks, include @mabdellatif88 on the request for review. If he should be tagged for secondary review, include a review task for the review to tag him after first-round review is complete._

- [ ] Check 1
- [ ] Check 2

PR #373, opened 2026-06-08
Fix YAML syntax for otel_collector and beyla staging-only roles

### LINKED ISSUE #372: Fix YAML syntax error for otel_collector and beyla staging-only roles
Commit [44a9998](https://github.com/Princeton-CDH/cdh-ansible/commit/44a999855cac236cca0b5122685a600e281f6976) introduced a YAML syntax error in `playbooks/escriptorium.yml` which causes Ansible to fail to parse the playbook entirely.

tanhaow:
**Associated Issue(s):** resolves Princeton-CDH/cdh-ansible#372

### Changes in this PR
- Fix YAML syntax error in `escriptorium.yml` where `when` conditions for `otel_collector` and `beyla` roles were placed at the wrong indentation level
- Use the role dict form so `when` is properly nested under each role entry

### Notes
- This bug was introduced in commit `44a9998` (pushed directly to main), which repeated the same mistake from PR #369 that had already been reverted once.
- The misplaced `when` caused Ansible to fail to parse the playbook entirely.

### Reviewer Checklist
- [ ] Review the code

### DIFF
--- a/playbooks/escriptorium.yml
+++ b/playbooks/escriptorium.yml
@@ -7,10 +7,10 @@
   remote_user: pulsys
   become: true
   roles:
-    - pulibrary.princeton_ansible.otel_collector
-  when: runtime_env == "staging"
-    - pulibrary.princeton_ansible.beyla
-  when: runtime_env == "staging"
+    - role: pulibrary.princeton_ansible.otel_collector
+      when: runtime_env == "staging"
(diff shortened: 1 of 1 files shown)


tanhaow:
Duplicate of #371. Noticed that it has already been fixed by @kayiwa in PR #371. Closing in favor of that PR.

PR #375, opened 2026-06-15
Fix activate alias in conan bash profile

tanhaow:
**Associated Issue(s):** resolves Princeton-CDH/htr2hpc#106

### Changes in this PR
- `bash_profile.j2`: use `{{ install_root }}/current` instead of `current_deploy` in the `activate` alias

### Reviewer Checklist
- [ ] Log in as `conan` on both VMs and verify the virtualenv is activated automatically

### DIFF
--- a/roles/deploy_user/templates/bash_profile.j2
+++ b/roles/deploy_user/templates/bash_profile.j2
@@ -4,7 +4,7 @@
 {% if single_app and install_root is defined %}
 # alias to change directory to current deployed version
 # and activate python virtualenv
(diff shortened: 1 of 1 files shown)


rlskoeser (changes requested):
@tanhaow This is the wrong fix - you need to address https://github.com/Princeton-CDH/htr2hpc/issues/105 

Once you fix the deploy as we agreed, I expect the existing alias to work correctly.

tanhaow:
Closing in favor of #376, which addresses the root cause by reverting to a single fixed deploy directory.

PR #376, opened 2026-06-15
Revert escriptorium deploy to single fixed directory

tanhaow:
**Associated Issue(s):** resolves Princeton-CDH/htr2hpc#105

### Changes in this PR
- `inventory/group_vars/htr/vars.yml`: revert `deploy` and `current_deploy` to `{{ install_root }}` (single fixed directory); remove `python_app_version` override

### Reviewer Checklist
- [x] Log in as `conan` on both VMs and verify the virtualenv is activated automatically

### DIFF
--- a/inventory/group_vars/htr/vars.yml
+++ b/inventory/group_vars/htr/vars.yml
@@ -20,6 +20,9 @@ gitref: dev-0.14.2
 repo: scripta/escriptorium
 # name of django application (used by build_project_repo)
 python_app: "escriptorium"
+# set version to match gitref since eScriptorium has no __version__ attribute;
+# build_project_repo will skip auto-detection when python_app_version is pre-set
+python_app_version: "{{ gitref }}"
 django_app: "{{ python_app }}"
 # django app & python requirements are nested under the app folder
 django_app_path: "{{ deploy }}/app"
--- a/inventory/group_vars/htr/vars.yml
+++ b/inventory/group_vars/htr/vars.yml
@@ -20,9 +20,6 @@ gitref: dev-0.14.2
 repo: scripta/escriptorium
 # name of django application (used by build_project_repo)
 python_app: "escriptorium"
-# set version to match gitref since eScriptorium has no __version__ attribute;
-# build_project_repo will skip auto-detection when python_app_version is pre-set
-python_app_version: "{{ gitref }}"
 django_app: "{{ python_app }}"
 # django app & python requirements are nested under the app folder
 django_app_path: "{{ deploy }}/app"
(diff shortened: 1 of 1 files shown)


rlskoeser (commented):
Great, glad this was a simple variable change.

It looks like the change has been made on cdh-htr2 based on the filesystem, but cdh-htr1 still has current and previous directories and I can tell from the website it's running the wrong version of htr2hpc.  Did you run it on both or only test it on one?

tanhaow:
> Great, glad this was a simple variable change.
> 
> It looks like the change has been made on cdh-htr2 based on the filesystem, but cdh-htr1 still has current and previous directories and I can tell from the website it's running the wrong version of htr2hpc. Did you run it on both or only test it on one?

Ah, I actually ran it on `htr1` first and verified that `conan` is no longer broken and points to the correct directory (code and virtualenv are both at the fixed path `/srv/www/escriptorium/` directly). But `finalize_deploy` (the last role in the playbook) still creates current and previous symlinks. 

I think when you checked htr2, the playbook hadn't finished yet (sorry for kicking the review request a bit early), so `finalize_deploy` hadn't executed and the symlinks hadn't been created yet, which is why it appeared correct. 

Should I update `finalize_deploy` to skip symlink creation, or just accept the circular current symlink for now since it doesn't affect functionality? @rlskoeser

rlskoeser:
@tanhaow let's not introduce more changes, matching current staging with the empty/non-useful symlinks is fine. But do please clean up the dev-0.14 deploy that's still hanging around, which the previous symlink is pointing to.

tanhaow:
Thanks for the suggestion. I cleaned up the old deploys on both VMs. The only remaining items are the current circular symlink and the previous empty symlink. @rlskoeser

rlskoeser:
> Thanks for the suggestion. I cleaned up the old deploys on both VMs. The only remaining items are the current circular symlink and the previous empty symlink. @rlskoeser


Great. Deploy directory looks good now, but the web interface is running the wrong version of the htr2hpc code.  (Did you forget to specify the release branch?)  This will be a good test - please run again with the htr2hpc reinstall tag and correct branch so we can confirm it does what we expect.

FYI, I removed the previous symlink, we don't need it. (It's ok if it gets created again, but let's not have it pointing at something that isn't there.)

tanhaow:
> > Thanks for the suggestion. I cleaned up the old deploys on both VMs. The only remaining items are the current circular symlink and the previous empty symlink. @rlskoeser
> 
> Great. Deploy directory looks good now, but the web interface is running the wrong version of the htr2hpc code. (Did you forget to specify the release branch?) This will be a good test - please run again with the htr2hpc reinstall tag and correct branch so we can confirm it does what we expect.
> 
> FYI, I removed the previous symlink, we don't need it. (It's ok if it gets created again, but let's not have it pointing at something that isn't there.)

Good catch! Have rerun it with the reinstall tag and the release branch.

rlskoeser (approved):
Looks right on both VMs and on the web interface. Let's go!

PR #381, opened 2026-06-24
I380 rclone

### LINKED ISSUE #380: setup role fails when downloading rclone because `rclone_download_dest` is undefined
The setup role fails during the rclone installation task because the variable rclone_download_dest is not defined.

### error

```sh
TASK [setup : Setup - download the latest Rclone binary] [ERROR]: Task failed: Finalization of task args for 'ansible.builtin.get_url' failed: Error while resolving value for 'dest': 'rclone_download_dest' is undefined
```

The failure occurs in:

roles/setup/tasks/rclone.yml:11

Specifically:

```yaml
- name: Setup - download the latest Rclone binary
  ansible.builtin.get_url:
    url: "{{ rclone_download_url }}"
    dest: "{{ rclone_download_dest }}"
```
### Affected hosts
cdh-test-shxco1.princeton.edu
cdh-test-shxco2.princeton.edu

### Impact

The playbook cannot complete the setup role on affected hosts. The failure happens before get_url runs because Ansible cannot render the task arguments.

### Expected behavior

The role should define a default download destination for the rclone archive, or the task should use an existing variable that is already defined for the role.

### Actual behavior

The role references `rclone_download_dest`, but that variable is undefined in the current inventory/group vars/defaults.

### Acceptance criteria

- `rclone_download_dest` is defined in role defaults or inventory.
- The rclone download task no longer fails during argument finalization.
- The setup role runs successfully on `cdh-test-shxco1.princeton.edu` and `cdh-test-shxco2.princeton.edu`.
- Any Molecule or CI coverage for the setup role is updated if applicable.

kayiwa:
**Associated Issue(s):** resolves #380 

### Changes in this PR

- Adds missing rclone install defaults used by the setup role:
  - `rclone_version`
  - `rclone_download_url`
  - `rclone_download_dest`
  - `rclone_extract_path`
- Adds missing rclone config path defaults:
  - `rclone_user_dir`
  - `rclone_config_dir`
  - `rclone_config_file`
  - `rclone_gdrive_credentials_file`
- Ensures the rclone config directory exists before writing remote configuration options.
- Fixes failures when `rclone_remotes` is configured for SHXCO hosts.

### Notes

- The role previously documented that `rclone_remotes` triggers rclone setup, but the tasks also required several path variables that were not defined.
- Rclone is pinned by version instead of relying on an implicit latest download path, making the role more repeatable.

### Reviewer Checklist

- [x] Confirm the rclone defaults are appropriate for SHXCO hosts.
- [x] Confirm the generated config path resolves to `/home/conan/.config/rclone/rclone.conf`.
- [x] Confirm the Google Drive service account file path resolves under the rclone config directory.
- [x] Confirm the playbook no longer fails on undefined rclone variables.


inventory/group_vars/shxco_staging/vars.yml
```diff
@@ -5,6 +5,8 @@
 
 running_on_server: true
 
+webpack_stats_check_enabled: false
```
rlskoeser: I think this is part of the other clean up pr - but I know it was needed for this deploy to run. At a minimum I'd like to move to all shxco vars but I wonder if it could be off by default for all hosts since I think only a few need it.

roles/setup/defaults/main.yml
```diff
@@ -1,3 +1,15 @@
...
+rclone_download_url: "https://downloads.rclone.org/v{{ rclone_version }}/rclone-v{{ rclone_version }}-linux-amd64.zip"
+rclone_download_dest: "/tmp/rclone-v{{ rclone_version }}-linux-amd64.zip"
+rclone_extract_path: "/tmp/rclone-v{{ rclone_version }}-linux-amd64"
+rclone_user: "conan"
+
+# rclone config defaults
+rclone_user_dir: "/home/{{ rclone_user }}"
+rclone_config_dir: "{{ rclone_user_dir }}/.config/rclone"
+rclone_config_file: "{{ rclone_config_dir }}/rclone.conf"
+rclone_gdrive_credentials_file: "{{ rclone_config_dir }}/pul-gcdc-02ad5e9d8029.json"
```
rlskoeser, line 15: This filename seems very specific - before I was calling it `gdrive_creds.json` 

I don't think it matters, just curious where it came from.

rlskoeser (commented):
Thanks for making the issue and opening the PR to resolve - I've been wondering how this got broken.

I went looking in the history: this is the last version of rclone vars previously:
https://github.com/Princeton-CDH/cdh-ansible/commit/7ff26b3fd18aa5073d9eb8575097067d5622a3ab 

It was removed in this commit:
https://github.com/Princeton-CDH/cdh-ansible/commit/ece131caf0ac73d3e346297e737d54ffcba88690

Unintentional, I presume - part of a big pr with lots of changes, so harder to notice that it was removed.

I can't find where the download urls and dest were set before, or if we were inheriting/assuming something from prancible that isn't there anymore.  Thanks for adding them.

Once I figured out the command, I confirmed that access to google drive is working through rclone - I can get a list of files on the staging server with this:
`rclone ls shxco_dataexports_drive:/`

rlskoeser:
@kayiwa the rclone improvements are great, my questions are around the webpack check stuff. Let me know how you'd prefer to resolve that - should I move the opt-out to all shxco hosts for now? Should we merge this and deal with that on the other pr?

roles/setup/defaults/main.yml
kayiwa: I didn't know where to find said file so I used an id-generator to make it pass.

rlskoeser (approved):
Thank you for fixing rclone

tanhaow (approved):

PR #389, opened 2026-07-06
Refactor webpack build into nodejs role

### LINKED ISSUE #344: The check for webpack status assumes a filepath that not all applications use
Currently it assumes the webpack status file is always under `/sitemedia`

Test if we need to keep this:
- [ ] check if geniza requires webpack stats to be web accessible (RSK)
- if not:
    - [ ] update geniza to change the webpack path (RSK, review by BS or MA)
    - [ ] remove this change in cdh-ansible (HT)
- if we need to keep it:
    - [ ] add a variable for webpack path so we can differentiate apps that need it web accessible (off by default) [HT]

### LINKED ISSUE #387: Convert webpack role to a task within node/npm role and include in molecule tests
Our currently `run_webpack` role is problematic and not well tested; it should really be a task within a role.  Let's fix it.

- convert to a task and move it into the appropriate role (probably `build_npm` but lets rename to something more meaningful (perhaps `nodejs` as suggested ages ago in #165 )
- update molecule for the role to test the webpack task
- clean up & improve debugging / verification as feasible
- resolve webpack check files and conventions - will resolve #344 
- update all playbooks that use npm and webpack to use the new role & task
- test all updated playbooks against staging (in ansible workflow, if we can)

kayiwa:
**Associated Issue(s):** resolves #387; resolves #344 

### Changes in this PR

- Adds a new `nodejs` role that consolidates Node.js installation, npm dependency installation, and webpack asset building.
- Moves the old `run_webpack` behavior into `roles/nodejs/tasks/webpack.yml` so webpack can be run as a task within the appropriate role.
- Adds `roles/nodejs/tasks/npm.yml` for npm dependency installation, replacing the older `build_npm` workflow.
- Adds configurable webpack build command mapping through `nodejs_webpack_build_commands`.
- Adds configurable webpack stats-file validation through `nodejs_webpack_stats_file` and `nodejs_webpack_stats_check`.
- Improves webpack debugging output by reporting the npm command, working directory, stdout, stderr, expected stats file, and candidate stats files found on disk.
- Adds Molecule coverage for the new role, including a fixture webpack application that builds assets and verifies generated webpack stats JSON.
- Adds Node.js install verification for the expected Node version and `/usr/local/bin` symlinks.
- Adds README documentation for role usage, variables, webpack stats conventions, debugging, Molecule testing, and migration from `build_npm / run_webpack`.

- 

### Notes

- This role keeps legacy variable fallbacks such as `desired_nodejs_version`, `npm_install_path`, `npm_install_mode`, `webpack_build_qa`, and `webpack_build_prod` to reduce migration risk.
- Applications using the default `Django / webpack-bundle-tracker` convention can continue using `sitemedia/webpack-stats.json`.
- Applications that write webpack stats somewhere else should set `nodejs_webpack_stats_file`.
- Applications that do not produce webpack stats should set `nodejs_webpack_stats_check: false`.
- This should help resolve the convention mismatch described in #344.
- Because this changes shared npm / webpack behavior across playbooks, reviewers should pay close attention to application-specific group vars and staging deploy behavior.

### Reviewer Checklist

- [x] Confirm roles/nodejs/tasks/main.yml includes the install, npm, and webpack task files with the expected conditionals.
- [x] Run molecule test from roles/nodejs and confirm the fixture webpack build passes.
- [x] Confirm the Molecule verify step validates Node.js, npm/corepack symlinks, and webpack stats JSON.
- [x] Confirm each migrated application has the correct nodejs_npm_install_path.
- [x] Confirm each migrated application has the correct webpack build command for its runtime_env.
- [x] Confirm applications with non-standard webpack stats paths set nodejs_webpack_stats_file.
- [x] Confirm applications without webpack stats output set nodejs_webpack_stats_check: false.
- [x] Test updated staging playbooks through the Ansible workflow where feasible.


kayiwa:
This successfully ran the cdhweb playbook on a jammy (cdh-test-web2) and noble (cdh-test-web1)

inventory/group_vars/cdhweb_staging/vars.yml
```diff
@@ -20,3 +20,8 @@ passenger_server_name: "{{ application_url }}"
 # observability
 beyla_enabled: true
 beyla_service_name: cdh-web
+# nodejs
+nodejs_version: "v22.4.0"
+nodejs_npm_enabled: true
+nodejs_webpack_enabled: true
+nodejs_webpack_stats_check: false
```
rlskoeser: Please move these to `group_vars/cdhweb/vars.yml` so both prod and staging can use the same config

playbooks/cdhweb.yml
```diff
@@ -13,18 +13,15 @@
     - role: pulibrary.princeton_ansible.beyla
       when: runtime_env | default('staging') == "staging"
 
-    - role: pulibrary.princeton_ansible.otel_collector
-      when: runtime_env | default('staging') == "staging"
```
rlskoeser, line 17: Why is this being removed? Doesn't seem relevant to the other changes

roles/nodejs/meta/main.yml
```diff
@@ -0,0 +1,18 @@
+---
+galaxy_info:
+  role_name: nodejs
+  company: CDH University Library
```
rlskoeser: Maybe both?

```suggestion
  company: CDH @ Princeton / Princeton University Library
```

roles/nodejs/meta/main.yml
```diff
@@ -0,0 +1,18 @@
...
+  description: Nodejs Role
+  author: cdh
+
+  license: MIT
```
rlskoeser: fwiw, I thought our others are apache 2 but it looks like we have a mix now...

rlskoeser (commented):
There are merge conflicts with `main` related to the recent passenger changes - I'm not 100% confident how to resolve.

If you update from main and add an `argument_specs.yml` file for the new role, it should be included in the docs and we'll get a preview on this pr.

This doesn't remove the old `build_npm` role - do we wait until we've transitioned all the playbooks over to the new one before we remove? How can we keep track and make sure we don't forget to clean it up?  

For testing: should I rerun playbook for cdhweb? Should I update vars and playbook for one of the other apps and test it that way?  I could start that in a new branch chained from this one.

I tried running molecule locally and I can't get it to work. Pretty sure this is something to do with my setup, but I would like to resolve it so I can run molecule tests. I tried triaging with OpenCode + glm-5.1 and it claimed there were a few tmp dir configurations missing (which it added but did not solve the problem), and then said the real problem is that Docker Rosetta emulation isn't working on my machine.

Here's the beginning of the error:
> TASK [Prepare | Create build user] *********************************************
[ERROR]: Task failed: Failed to create temporary directory.

rlskoeser:
I made several changes in this branch related to the sphinx documentation & roles:
- update sphinx extension for roles so the role index will be updated when new argument_spec files are added
- improve notes in the readme for how to build the docs (including install dependencies for building docs)
- add a .justfile and add a note in the readme for convenient shortcut to build docs

I also made some adjustments to the node/webpack config for cdhweb:
- moved variables up so they're all in the same space, and reconciled versions
- updated the build mode (cdhweb requires ci rather than production because of how things are specified)  
- all the docs I found say cdhweb is on node v20 but it's actually using 22.4 on the servers and 20 is EOL, so I'm going to leave it as 22.4

Note: I think the cdhweb deploy succeeded previously because you were deploying a version that was already present on the server (our deploys create directories based on git commit hashes and there wasn't a new version to deploy). I tested installing a version that wasn't deployed to staging, got the error, adjusted the build command, and it succeeded.

rlskoeser:
@tanhao I worked on this with Francis last week and added some changes to make it easier to add new roles to the documentation.  (I can't run the molecule tests locally but they are passing in CI and we decided not to worry about that.)  I think it's in good shape, but would appreciate if could look this in tandem with #408, where I've switched shxco over to the new role.  My hope is that you would do the same for htr2hpc playbook in the next iteration or so, so ask or comment if there isn't enough context for you to do that.

roles/nodejs/tasks/webpack.yml
```diff
@@ -0,0 +1,150 @@
...
+      ansible.builtin.command:
+        argv:
+          - npm
+          - run
+          - "{{ nodejs_webpack_build_command }}"
+        chdir: "{{ nodejs_npm_install_path }}"
+      environment:
+        PATH: "{{ nodejs_npm_install_path }}/node_modules/.bin:/usr/local/bin:/usr/bin:/bin"
+      register: nodejs_webpack_result
+      changed_when: true
```
tanhaow: Should be `changed_when: false`. With true, this task always reports changed, right? which would cause the molecule idempotence check to fail.

roles/nodejs/molecule/default/ansible.cfg
```diff
@@ -0,0 +1,2 @@
+[defaults]
+remote_tmp = /tmp/ansible
```
tanhaow, line 2:
```suggestion
remote_tmp = /tmp/ansible
roles_path = ../../../:~/.ansible/roles:/usr/share/ansible/roles:/etc/ansible/roles
```

Local molecule test fails at converge because Ansible can't find the nodejs role. It runs from inside `roles/nodejs/` and never looks up at the repo-level `roles/` directory. It needs a `roles_path` pointing up three levels.

roles/nodejs/molecule/default/molecule.yml
```diff
@@ -0,0 +1,41 @@
...
+      - /sys/fs/cgroup:/sys/fs/cgroup:rw
+    cgroupns_mode: host
+    privileged: true
+    pre_build_image: true
+provisioner:
+  name: ansible
+  playbooks:
+    create: create.yml
+    destroy: destroy.yml
+  log: true
```
tanhaow, line 42: Same as above, also needs a `roles_path` defined. Otherwise the module test is failing

tanhaow (changes requested):
My local molecule test fails at converge because Ansible can't find the nodejs role.

roles/nodejs/molecule/default/molecule.yml
kayiwa: Good catch. I couldn’t reproduce the role lookup failure locally, but I agree the Molecule scenario should not depend on my cwd or local Ansible config. I added an explicit roles_path so the scenario can resolve nodejs consistently.

I also changed the webpack task to changed_when: false. Since Molecule runs converge again during idempotence, forcing that task to report changed would make idempotence fail even when the build succeeds.

roles/nodejs/molecule/default/molecule.yml
```diff
@@ -0,0 +1,44 @@
...
+    privileged: true
+    pre_build_image: true
+provisioner:
+  name: ansible
+  playbooks:
+    create: create.yml
+    destroy: destroy.yml
+  config_options:
+    defaults:
+      roles_path: ../../..
```
tanhaow:
```suggestion
      roles_path: ${MOLECULE_PROJECT_DIRECTORY}/..:~/.ansible/roles:/usr/share/ansible/roles:/etc/ansible/roles
```
I ran test locally and it still failed. The error message returns `the role 'nodejs' was not found`.

I believe the issue is caused by here:

`roles_path: ../../..` works in `ansible.cfg` because Ansible resolves it relative to the file's own location (`roles/nodejs/molecule/default/` - go up 3 levels), but not here because Molecule creates a temp working dir and run within it. So we need to pass in an environment variable (`{MOLECULE_PROJECT_DIRECTORY}`) to make it an absolute path.

tanhaow (changes requested):
Please see my comment below.

tanhaow (approved):
Thanks for your revision! The local molecule test all passed on my end 🚀

PR #391, opened 2026-07-08
Fix/hpc hostname ansible variable

tanhaow:
**Associated Issue(s):** resolves Princeton-CDH/htr2hpc#88

### Changes in this PR

- Move hardcoded `HPC_HOSTNAME` value in `escriptorium_settings.py.j2` to an ansible variable `hpc_hostname`

### Reviewer Checklist

- [ ] Confirm `HPC_HOSTNAME` is correctly rendered in the deployed settings file

### DIFF
--- a/inventory/group_vars/htr/vars.yml
+++ b/inventory/group_vars/htr/vars.yml
@@ -43,6 +43,7 @@ node_version: "18"
 # not the htr2hpc package. To deploy a specific htr2hpc branch or tag, use:
 #   -e htr2hpc_gitref=adroit
 htr2hpc_gitref: main
+hpc_hostname: "adroit.princeton.edu"
 python_extra_packages:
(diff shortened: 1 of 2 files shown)


rlskoeser (approved):
Change looks like what I would expect. 

Your review notes says to check deployed settings, have you run the deploy with this setting change to update staging? And will this be a change from what is already there so we can tell if there's any difference?  

We could treat checking as part of acceptance testing rather than review, if that's helpful.

tanhaow:
> Change looks like what I would expect.
> 
> Your review notes says to check deployed settings, have you run the deploy with this setting change to update staging? And will this be a change from what is already there so we can tell if there's any difference?
> 
> We could treat checking as part of acceptance testing rather than review, if that's helpful.

Since this change doesn't affect the value of HPC_HOSTNAME, I believe there won't be a visible difference. But the value can be confirmed by checking the deployed settings file directly. I've uodated the acceptance notes on the issue!

PR #397, opened 2026-07-10
Fixes for postgres apt and SSH host key issues

### LINKED ISSUE #356: Fix postgres apt repository to include signed-by option
The `postgresql` role's default `postgres_apt_repository` value does not include the `signed-by` option:

```
deb http://apt.postgresql.org/pub/repos/apt/ ...
```

On Ubuntu 22.04 (jammy), if the server already has a `signed-by` entry for the same repo (written by a previous deploy), apt will fail with:

```
Conflicting values set for option Signed-By
```

**Current workaround:** Override `postgres_apt_repository` in `inventory/group_vars/htr/vars.yml` to include `signed-by`.

**Fix:** Update the default value in `roles/postgresql/defaults/main.yml` to include `signed-by`, so no per-project override is needed.

blms:
**Associated Issue(s):** resolves #356

### Changes in this PR

- Resolves #356 by adding a step to remove the old PDPG source files for psql
- Resolves issue with the app server's SSH host key changing from what's stored on the solr server.
   - The existing `-o StrictHostKeyChecking=no` partially fixed this by auto-accepting unknown hosts, but it didn't work for changed ones (e.g. after a VM rebuild). `-o UserKnownHostsFile=/dev/null` just simply ignores known_hosts when performing the rsync step. 
   - If that's not desired, we should probably add troubleshooting steps to the README instead about how to manually update known_hosts to remove and accept a changed key.
- Unrelated, but just while I'm in here, I resolved the deprecation warnings about using `port` instead of `login_port` per [ansible docs](https://docs.ansible.com/projects/ansible/latest/collections/community/postgresql/postgresql_user_module.html#parameter-login_port)

### Notes

- Used Claude to help with diagnosing root causes and making these changes.
- I was able to do a successful geniza staging deploy from this branch: https://ansible-tower.princeton.edu/#/jobs/playbook/52668/output
   - FWIW, I did have to manually go in and install  `passenger` and `libnginx-mod-http-passenger` from apt on both test servers (after [this failure](https://ansible-tower.princeton.edu/#/jobs/playbook/52664/output)), but I'm pretty sure that would be covered by running with the `setup` tasks enabled.
   - I had to skip the `gh_deploy` tag again, as otherwise I'm getting PAT errors. Here's a [failed run with the PAT error](https://ansible-tower.princeton.edu/#/jobs/playbook/52652/output) if someone wants to take a look. It looks like it's not expired, but needs different permissions to create the deploy.

### DIFF
--- a/roles/postgresql/tasks/backup_db.yml
+++ b/roles/postgresql/tasks/backup_db.yml
@@ -20,7 +20,7 @@
       community.postgresql.postgresql_db:
         name: "{{ application_db_name }}"
         login_host: "{{ postgres_host }}"
-        port: "{{ postgres_port }}"
+        login_port: "{{ postgres_port }}"
         login_user: "{{ application_dbuser_name }}"
         login_password: "{{ application_dbuser_password }}"
         encoding: "UTF-8"
--- a/roles/postgresql/tasks/create_db.yml
+++ b/roles/postgresql/tasks/create_db.yml
@@ -7,7 +7,7 @@
   community.postgresql.postgresql_db:
     name: "{{ application_db_name }}"
     login_host: "{{ postgres_host }}"
-    port: "{{ postgres_port }}"
+    login_port: "{{ postgres_port }}"
     login_user: "{{ postgres_admin_user }}"
     login_password: "{{ vault_postgres_admin_password }}"
     encoding: "UTF-8"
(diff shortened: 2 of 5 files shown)


kayiwa (approved):
OK. Thanks @blms

PR #402, opened 2026-07-14
Remove recurse from deploy directory task

### LINKED ISSUE #392: playbook deploy fails with 'maximum recursion depth exceeded' on VMs with large file counts
The `Create the specific deploy directory` task in uses `recurse: yes` -- calling this task wwhen deploying eScriptorium made Ansible to walk the entire `/srv/www/escriptorium` directory tree to set ownership and permissions. With ~177k files on eScriptoiurm VMs, this exceeds Python's recursion limit and fails the deploy.

I discovered this issue a while ago when deploying eScriptorium. I made the change locally to ensure the deploy playbook runs successfully, but I forgot to report it. Remove `recurse: yes` would fix this issue.

tanhaow:
Closes #392

### What's changed:
- Removes `recurse: yes` from the `Create the specific deploy directory` task in `build_project_repo`
- On eScriptorium VMs with ~177k files, recursively setting ownership exceeds Python's recursion limit and fails the deploy

### Test plan
- [ ] Run `playbooks/escriptorium.yml` against staging and confirm the deploy directory task completes without recursion error

🤖 Assisted-by: OpenCode:claude-opus-4.7 (for identifying the problem)

### DIFF
--- a/inventory/all_hosts
+++ b/inventory/all_hosts
@@ -102,8 +102,6 @@ cdh-htr2.lib.princeton.edu
 [htr_staging]
 cdh-test-htr1.lib.princeton.edu
 cdh-test-htr2.lib.princeton.edu
-cdh-test-legacy-htr1.lib.princeton.edu
-cdh-test-legacy-htr2.lib.princeton.edu
 
 [htr:children]
 htr_production
--- a/roles/build_project_repo/tasks/main.yml
+++ b/roles/build_project_repo/tasks/main.yml
@@ -47,7 +47,6 @@
         owner: "{{ deploy_user }}"
         group: "{{ deploy_user }}"
         mode: '0755'
-        recurse: yes
       tags: [setup]
 
     - name: Mark deploy path as safe for git


rlskoeser (commented):
I think you're right that we should remove this but what bothers me is I don't remember why we added it and I'm not sure what might break.  If anything breaks we should fix it a different way (probably fix at file/directory creation time), but it would be good to know now.

Your review notes say to test deploying to escriptorium staging. I think you should do that (if you have not already).  I could test deploying s&co and we could ask Mohamed to check with geniza.  What do you think?

tanhaow:
> Your review notes say to test deploying to escriptorium staging. I think you should do that (if you have not already). I could test deploying s&co and we could ask Mohamed to check with geniza. What do you think?

Yes I have tested it with deploying eScriptorium to staging. After removing recurse the playbook runs without problem. I think it's a good idea to have you test on S@Co and Mohamed test on Geniza. Tagging him now, @mabdellatif88

kayiwa (dismissed):
OK! Thanks @tanhaow

rlskoeser (dismissed):
Just tested deploying s&co to staging from this branch, no problems.

mabdellatif88 (commented):
Geniza staging deploy (of the PR's cdh-ansible branch, using ansible tower) is failing on [cdh-test-geniza2.princeton.edu](http://cdh-test-geniza2.princeton.edu/) (and seems to be passing on cdh-test-geniza1) citing "Resource not accessible by personal access token" . However, it is showing 0 recursion on both hosts. Screenshot of the full/detailed error attached

Image

mabdellatif88 (dismissed):

PR #405, opened 2026-07-21
Fix reinstall-htr2hpc tag: use pip executable instead of virtualenv

tanhaow:
### What's changed

- Replace `virtualenv: "{{ django_venv_path }}"` (no `virtualenv_command`) with `virtualenv: "{{ django_venv_path }}"` + `virtualenv_command: "python{{ python_version }} -m venv"` in the `reinstall-htr2hpc` tasks, matching the pattern used in the `python` role

### Notes

The `reinstall-htr2hpc` tasks were using `virtualenv:` without `virtualenv_command`, which causes Ansible's `pip` module to look for a standalone `virtualenv` executable. The HTR servers don't have `virtualenv` installed — they use Python's built-in `venv` module. Other roles (e.g. `python/tasks/install_dependencies.yml`) handle this correctly by also specifying `virtualenv_command: python{{ python_version }} -m venv`. This tag was newly added and had never been run before, so the issue wasn't caught earlier.

The `django_venv_path` variable is fine for HTR — `htr/vars.yml` defines `deploy: "{{ install_root }}"` (single-directory deploy, no hash-based paths), so `django_venv_path` correctly resolves to `/srv/www/escriptorium/app/env`.

### DIFF
--- a/roles/escriptorium_setup/tasks/main.yml
+++ b/roles/escriptorium_setup/tasks/main.yml
@@ -77,13 +77,13 @@
     # remove the current version
     - name: uninstall htr2hpc python package
       ansible.builtin.pip:
-        virtualenv: "{{ django_venv_path }}"
+        executable: "{{ install_root }}/app/env/bin/pip"
         name: htr2hpc
         state: absent
     # reinstall with whatever version is in settings
     - name: install currently configured htr2hpc python package
       ansible.builtin.pip:
-        virtualenv: "{{ django_venv_path }}"
+        executable: "{{ install_root }}/app/env/bin/pip"
         name: "{{ python_extra_packages }}"
         extra_args: "{{ python_extra_packages_pip_extra_args }}"
       notify:
--- a/roles/escriptorium_setup/tasks/main.yml
+++ b/roles/escriptorium_setup/tasks/main.yml
@@ -77,13 +77,15 @@
     # remove the current version
     - name: uninstall htr2hpc python package
       ansible.builtin.pip:
-        executable: "{{ install_root }}/app/env/bin/pip"
+        virtualenv: "{{ django_venv_path }}"
+        virtualenv_command: "python{{ python_version }} -m venv"
         name: htr2hpc
         state: absent
     # reinstall with whatever version is in settings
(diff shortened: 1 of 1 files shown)


rlskoeser:
Thank you for documenting  the problems so clearly.  Some questions:

> 1. `virtualenv:` syntax requires the `virtualenv` tool, but our current HTR VMs use Python's built-in `venv` and don't have `virtualenv` installed, so Ansible fails with "Failed to find required executable virtualenv". Now using `executable:` points directly to the pip inside the existing venv, so no extra tools needed.

I'm a little surprised that virtualenv syntax doesn't work with built-in `venv` module - it seems like that would have implications for all of our python deploys. Wasn't this working before? What changed?

> 2. `django_venv_path` expands to `{{ deploy }}/app/env`, and `deploy` is `install_root/version-shorthash` computed at runtime. When running only `reinstall-htr2hpc` without a full deploy, the hash is recalculated from the current clone and points to a directory that doesn't exist. Now changing it to `{{ install_root }}/app/env/bin/pip` to point to the stable venv location that doesn't change between deploys.

I know that the version-shorthash problem impacts other playbooks when we try to run with only specific tags or update existing deploys instead of current. But for escriptorium, aren't we still using a single-directory deploy? I would expect this to be the one project where the hash-based directory wouldn't come into play.

tanhaow:
@rlskoeser Good catches. You're right.

I know that eScriptorium does use a single fixed directory deploy. But I was somehow misled by an old deploy with shorthash on staging. I'll manually remove that.

For the `virtualenv:` syntax, I checked our other ansible playbooks and found that the python role always pairs it with a `virtualenv_command` (`virtualenv_command: python{{ python_version }} -m venv`), which tells Ansible to use the built-in venv module. The `reinstall-htr2hpc` task was missing the `virtualenv_command` part. I've updated the PR to follow the same pattern.

rlskoeser:
I'd forgotten about `virtualenv_command`! Thanks for finding that and explaining - makes sense. (Would love it if we could switch everything over to `uv`.... But that's a different project! 😅 )

rlskoeser (approved):

PR #407, opened 2026-07-24
Remove lingering datadog references and configurations

rlskoeser:
**Associated Issue(s):** N/A

### Changes in this PR

- Remove lingering datadog configurations in group vars / all vars
- Remove datadog role reference in prosody playbook

# Notes

- We previously removed datadog task in #273


kayiwa (approved):
OK! Thanks @rlskoeser

PR #417, opened 2026-08-06
Apply ESC-2026 security patch to eScriptorium on deploy

tanhaow:
related issue: https://github.com/Princeton-CDH/htr2hpc/issues/143

### What's changed

- Applies the ESC-2026 security patch to the Princeton eScriptorium deployment
- Fixes a bug encountered during staging validation.

#### Apply ESC-2026 security patch (`escriptorium_setup` role)
- Adds a vaulted patch file containing 8 security commits
- Copies the patch to the remote, applies it against the deploy directory, then removes it
- Patch is applied **before** other `replace` tasks to avoid context mismatch failures caused by earlier tasks modifying the same files
- Removes `fetch.py` and `test_fetch.py` before applying, because `patch` is not idempotent for new-file hunks — if those files already exist from a previous partial deploy, `patch` exits with an error instead of skipping

#### Python version pinning (`python` role)
- Staging was found to be running Python `3.11.0rc1`, installed by a previous setup run where the Princeton math mirror had a higher apt priority (600) than deadsnakes (500), causing the `rc1` pre-release on the Princeton math mirror to win over the deadsnakes `3.11.15` formal release
- The `3.11.0rc1` build lacks `sys.get_int_max_str_digits`, which torch/kraken imports unconditionally, causing a startup crash
- Fixes by adding an apt preferences file that pins `python3.11*` packages to the deadsnakes PPA at priority 900, ensuring the formal release is always preferred regardless of Princeton mirror priorities

## Reviewer's Checklist
- [ ] Review the code


roles/escriptorium_setup/tasks/main.yml
rlskoeser: Didn't you already fix these somewhere else?

roles/escriptorium_setup/tasks/main.yml
```diff
@@ -17,6 +17,33 @@
     # because we don't want to make it any easier than necessary
     # to ssh into other users accounts
 
+    - name: Remove files that the ESC-2026 patch will create (patch is not idempotent for new files)
+      ansible.builtin.file:
+        path: "{{ deploy }}/{{ item }}"
+        state: absent
+      loop:
+        - app/apps/imports/fetch.py
+        - app/apps/imports/tests/test_fetch.py
```
rlskoeser: interesting - thanks for documenting clearly why this is needed

roles/python/tasks/install_python.yml
```diff
@@ -14,6 +14,16 @@
...
 
+- name: Pin python packages to deadsnakes to avoid rc/alpha versions from distro mirror
+  become: true
+  ansible.builtin.copy:
+    dest: /etc/apt/preferences.d/deadsnakes-python
+    content: |
+      Package: python{{ python_version }} python{{ python_version }}-* libpython{{ python_version }}*
+      Pin: release o=LP-PPA-deadsnakes
+      Pin-Priority: 900
+    mode: "0644"
```
rlskoeser, line 30: Wild that this is needed! any idea if this affected any of the other htr2hpc VMs?  With ubuntu packages we can't specify more granular versions, can we?

roles/escriptorium_setup/tasks/main.yml
```diff
@@ -17,6 +17,33 @@
     # because we don't want to make it any easier than necessary
     # to ssh into other users accounts
 
+    - name: Remove files that the ESC-2026 patch will create (patch is not idempotent for new files)
```
rlskoeser: Can the new set of tasks to apply the patch be done as a block? (I forget what context blocks are allowed / disallowed). Or maybe a separate task?  (my goal is to make it easy to skip / turn off / remove)

I think it's important to have documentation or comments somewhere - maybe here - what this is for, so that it's easy to remove later one once we no longer need it, when we upgrade to a newer version that includes the security patches.

rlskoeser (dismissed):
Changes look good to me - thanks for figuring this out.

This seems good to go, but please add documentation somewhere about what these steps are for / why they are needed so it's easy to revisit later on (since we expect it to be temporary).

roles/python/tasks/install_python.yml
tanhaow: It took me some time to dig into this... So, both production and staging have this problem, but only staging has exposed it so far, because we rebuilt the venv from scratch on new staging, which triggered a fresh apt install python3.11. Production (the old staging) has never had its venv rebuilt, so the issue was never triggered there.

For more granular version pinning: Yep, Ubuntu's apt doesn't support pinning to a specific patch version. Packages are identified by minor version only (python3.11). So here I pinned the source to deadsnakes PPA at priority 900 instead to win over the Princeton mirror, to maje sure it uses the formal release 3.11.15. Do you know of a better approach? And should I report this to the #cdh-infra?

roles/escriptorium_setup/tasks/main.yml
tanhaow: Good suggestion. I wrapped the four patch-related tasks in a block so it's easier to remove them as a unit. I've also added the comments.

tanhaow:
Merge this branch as it has previously been approved, and the new revisions are just trivial and comment additions.

